PR #64, opened 2021-10-01
Saad/feature/connect popup app launch support

saadibrahim:
Screen Shot 2021-10-01 at 9 16 05 PM

Desktop is pretty much the same, without the "click to launch verida app" link.

Screen Shot 2021-10-01 at 9 13 15 PM

We need to work on design for mobile.

I considered only using media queries to switch content based on device type but then realized if someone was using desktop browser in lower width they would see the mobile version so went with a javascript (detecting user agent) + css solution.


saadibrahim:
Mobile design updated.

Screen Shot 2021-10-06 at 7 10 42 PM

Issue #70, opened 2021-10-08
Create API documentation


tahpot:

tahpot:
Decided to use the [better-docs](https://github.com/SoftwareBrothers/better-docs) theme as it looks nice and supports typescript.

Here's a good [intro article](https://wojciechkrysiak.medium.com/typescript-jsdoc-better-docs-7c03b6ea04df).

tahpot:
@nick-verida I've got this building the docs, however for some reason it only includes some files and not others.

I've run out of time today, so handing over to you to have a look.

PR #85, opened 2021-10-26
Feature/70 create api docs

cmcWebCode40:
- Modified `jsdoc.json` config file

resolves issue #70


tahpot:
@cmcWebCode40 I've run this locally, but it appears lots of the classes are still not available.

For example:

- https://github.com/verida/verida-js/blob/main/packages/client-ts/src/context/engines/verida/messaging/engine.ts
- https://github.com/verida/verida-js/blob/main/packages/client-ts/src/context/engines/verida/messaging/inbox.ts
- https://github.com/verida/verida-js/blob/main/packages/client-ts/src/context/engines/verida/messaging/outbox.ts
- https://github.com/verida/verida-js/blob/main/packages/client-ts/src/context/engines/verida/database/db-encrypted.ts

cmcWebCode40:
> * Modified `jsdoc.json` config file
> 
> resolves issue #70

@tahpot, I'm currently having issues displaying all the classes, it does not display more than 7 classes, I have also exported all default class names correctly. have you experienced issues like this working with  `better-docs`  ?.

tahpot:
@cmcWebCode40 I have never used better-docs sorry. I chose it very quickly as it looked like the nicest looking solution that supported `typescript`.

@nick-verida Is it worth us looking at a different library?

nick-verida:
@cmcWebCode40 do you know where this is at? Is it waiting for @tahpot to take another look at it?

tahpot:
@nick-verida @cmcWebCode40 

I have merged this in as the core requirement of getting the API docs is working.

It still needs more complete API docs added to the code. I have created a task here https://github.com/verida/verida-js/issues/114

PR #126, opened 2022-01-13
Add notification server support

pkhien95:
Add typing for account-web-vault to support notification server


nick-verida (commented):
Looks good to me. I'll leave to @tahpot to approve

PR #128, opened 2022-01-18
passed config to ping

kalanamt:


packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -27,15 +28,18 @@ export default class NotificationEngineVerida implements BaseNotification {
...
-    public async ping(): Promise<boolean> {
-        await this.init()
-        const server = await this.getAxios()
-
+    public async ping(config?: MessageSendConfig): Promise<boolean> {
+        await this.init();
+        const server = await this.getAxios();
+        
+        let context = config ? config.recipientContextName || this.context.getContextName() : this.context.getContextName();
+        let did = config ? config.did || this.did! : this.did!;
```
pkhien95, line 36: config.did is `undefined`

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -27,15 +28,18 @@ export default class NotificationEngineVerida implements BaseNotification {
...
         try {
+            // Returns the client context and the corresponding `DID`
             await server.post(this.serverUrl + 'ping', {
```
pkhien95, line 39: when I test, 
```
this.serverUrl + 'ping'
```
always miss a '/' character. for example `https://verida.ioping` instead of `https://verida.io/ping`

kalanamt: @pkhien95 This will fix in a separate PR

kalanamt: Fixed

nick-verida:
from @pkhien95 

> @nick-verida Sorry I missed this when started this feature: currently the server is sending a `data` type of message, which will not trigger a push notification from OS when the app is in background. We have 2 ways to resolve this:
> 
>     * Send a `notification` type of message. Official doc here: https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#notification
> 
>     * Send a "wake-up" data message:
>       
>       * iOS: add `content-available: true`
>       * Android: add `priority: high`
> 
> 
> Let me know which way do you think we should use.

I'm not sure, but it sounds like the `wake-up` message matches what we want right? We want to wake the app up and get it to poll?

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -18,24 +19,31 @@ export default class NotificationEngineVerida implements BaseNotification {
...
-        const server = await this.getAxios()
+    public async ping(config?: MessageSendConfig, did?: any): Promise<boolean> {
+        await this.init();
+        const server = await this.getAxios();
+        
+        let context = config ? config.recipientContextName || this.context.getContextName() : this.context.getContextName();
+
+        this.serverUrl = this.serverUrl.endsWith("/") ? this.serverUrl : this.serverUrl + "/";
+
+        console.log(`verida-route:- ping \n config:- ${config} \n did:- ${did} \n context:- ${context}`);
```
tahpot, line 39: Remove console log before merging?

packages/client-ts/src/context/engines/verida/messaging/outbox.ts
```diff
@@ -68,10 +68,13 @@ class VeridaOutbox {
     did = did.toLowerCase();
 
     const defaults: MessageSendConfig = {
+      did: did,
       // By default send data to the user's official Verida Vault application
       recipientContextName: VAULT_CONTEXT_NAME,
       // @todo: set a default expiry that is configurable but defaults to 24 hours?
     };
+    // TODO: This replaces the `did` specified in config
```
tahpot, line 76: Refactor now or in a separate issue later?

packages/client-ts/src/context/engines/verida/messaging/engine.ts
```diff
@@ -74,7 +74,7 @@ class MessagingEngineVerida implements BaseMessage {
     // Ping the notification service if it exists
     // @todo: Make it configurable if the notification service is pinged
     if (response && this.notificationService) {
-      await this.notificationService.ping()
+      await this.notificationService.ping(config, did);
```
tahpot, line 77: Can we pass just the context name from the config instead of the whole config object? That way we minimise the interdependency between the different classes.

tahpot (changes requested):

packages/client-ts/src/context/engines/verida/messaging/engine.ts
kalanamt: @tahpot  The implemented `ping(config?:MessageSendConfig, ...)` Current clients reads the config object. This will limit the configuration we can pass here.

packages/client-ts/src/context/engines/verida/messaging/outbox.ts
kalanamt: Addresses in :- https://github.com/verida/verida-js/issues/131

kalanamt:
@tahpot  PR Updated

packages/client-ts/src/client.ts
```diff
@@ -48,6 +49,8 @@ class Client {
       : {};
     const config = _.merge(defaultConfig, userConfig);
 
+    this.defaultVaultAppName = DEFAULT_CONFIG.vaultAppName;
```
tahpot, line 52: Should come from `config`, otherwise user specified defaults won't be applied.

packages/client-ts/src/context/engines/verida/messaging/outbox.ts
```diff
@@ -68,10 +68,14 @@ class VeridaOutbox {
     did = did.toLowerCase();
 
     const defaults: MessageSendConfig = {
+      did: did,
       // By default send data to the user's official Verida Vault application
       recipientContextName: VAULT_CONTEXT_NAME,
       // @todo: set a default expiry that is configurable but defaults to 24 hours?
+      // Fix in :- https://github.com/verida/verida-js/issues/131.
     };
+    // TODO: This replaces the `did` specified in config
```
tahpot, line 77: Can these comments be removed or if not, more informative?

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -18,24 +19,27 @@ export default class NotificationEngineVerida implements BaseNotification {
...
       * Ping a notification server to fetch new messages
       */
-    public async ping(): Promise<boolean> {
-        await this.init()
-        const server = await this.getAxios()
+    public async ping(recipientContextName: string, did?: any): Promise<boolean> {
+        await this.init();
+        const server = await this.getAxios();
+       
+        this.serverUrl = this.serverUrl.endsWith("/") ? this.serverUrl : this.serverUrl + "/";
```
tahpot, line 35: Perhaps do this in the constructor?

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -18,24 +19,27 @@ export default class NotificationEngineVerida implements BaseNotification {
...
+    }
 
-     /**
+    /**
       * Ping a notification server to fetch new messages
       */
-    public async ping(): Promise<boolean> {
-        await this.init()
-        const server = await this.getAxios()
+    public async ping(recipientContextName: string, did?: any): Promise<boolean> {
```
tahpot, line 31: `did` should be required

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -18,24 +19,27 @@ export default class NotificationEngineVerida implements BaseNotification {
     /**
      * Initialize notifications for the connected user
      *
-     * (ie; connect to a notification server)
+     *    * (ie; connect to a notification server)
```
tahpot, line 22: This method can be removed along with `this.did`.

packages/client-ts/src/context/engines/verida/notification/engine.ts
```diff
@@ -18,24 +19,27 @@ export default class NotificationEngineVerida implements BaseNotification {
...
+       
+        this.serverUrl = this.serverUrl.endsWith("/") ? this.serverUrl : this.serverUrl + "/";
 
         try {
+            // Returns the client context and the corresponding `DID`
             await server.post(this.serverUrl + 'ping', {
                 data: {
-                    did: this.did!,
-                    context: this.context.getContextName()
+                    did: did ? did : this.did!,
```
tahpot, line 41: This is the heart of the actual bug.

The code should never have used `this.did` as that is the DID of the current user. The system should never send a notification to the current user, only to the user that the notification is being sent to.

As such, this should just be `did: did`.

packages/client-ts/src/context/interfaces.ts
```diff
@@ -113,6 +113,7 @@ export enum PermissionOptionsEnum {
  * Interface for any MessageSendConfig
  */
 export interface MessageSendConfig {
-  recipientContextName?: string;
+  did?: string,
```
tahpot, line 116: `did` shouldn't be optional

packages/client-ts/src/context/notification.ts
```diff
@@ -13,8 +14,9 @@ export default interface Notification {
 
     /**
      * Ping a notification server to fetch new messages
+     * TODO: Change this 
      */
-    ping(): Promise<boolean>
+    ping(recipientContextName? :string, did?: any): Promise<boolean>
```
tahpot, line 19: Both parameters need to be required.

tahpot (changes requested):
See review comments.

Lmk if anything needs clarification.

packages/client-ts/src/context/engines/verida/notification/engine.ts
tahpot: Perhaps rename `did` to `didToNotify` for greater clarity.

packages/client-ts/src/client.ts
kalanamt: the `config` does not have `defaultVaultAppName`

```javaScript
export interface ClientConfig {
	/**
	 * Environment to load by default.
	 *
	 * Environment configuration can still be overriden by config items.
	 */
	environment?: EnvironmentType;

	/**
	 * URL of Verida DID Server node to use.
	 */
	didServerUrl?: string;

	/**
	 * Specify custom schema paths (typicaly for local development).
	 */
	schemaPaths?: object;
}

```
Not sure  if its possible to read it from  `config`

packages/client-ts/src/context/engines/verida/notification/engine.ts
kalanamt: @tahpot  The `init` method cannot be reviewed since its been declared in Base Interface
```
export default interface Notification {

    /**
     * Initialize notifications for the connected user
     *
     * (ie; connect to a notification server)
     */
    init(): Promise<void>

    /**
     * Ping a notification server to fetch new messages
     * TODO: Change this 
     */
    ping(recipientContextName :string, didToNotify: any): Promise<boolean>

    getErrors(): string[]

}
```
and this is being referred in `class MessagingEngineVerida implements BaseMessage `. 

 Added an empty init method for `export default class NotificationEngineVerida implements BaseNotification ` 

This needs to be addressed separately since its gonna change the current implementation

pkhien95:
Please consider discussing my concern here: https://github.com/verida/vault-mobile/issues/219#issuecomment-1024058184

PR #129, opened 2022-01-20
Versionless schema signature generation

kalanamt:


kalanamt:
@tahpot  committing the work so far.

Issue #133, opened 2022-01-29
Issue with font loading on the Vault SSO SDK


attack-monkey:
Using Parcel as the Bundler for Web Application, I got the following error:

```

@parcel/core: No transformers found for
_Sites/pingle2/client/node_modules/@verida/account-web-vault/dist/assets/fonts/Sora-Regular.ttf_.

  /Users/benjaminclark/Sites/pingle2/client/node_modules/@parcel/config-default/index.json:3:3
     2 |   "bundler": "@parcel/bundler-default",
  >  3 |   "transformers": {
  >    |   ^^^^^^^^^^^^^^^^^
  >  4 |     "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >  5 |     "bundle-text:*": ["...", "@parcel/transformer-inline-string"],
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >  6 |     "data-url:*": ["...", "@parcel/transformer-inline-string"],
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >  7 |     "worklet:*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >  8 |       "@parcel/transformer-worklet",
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >  9 |       "..."
  >    | ^^^^^^^^^^^
  > 10 |     ],
  >    | ^^^^^^
  > 11 |     "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 12 |       "@parcel/transformer-babel",
  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 13 |       "@parcel/transformer-js",
[...]

PR #134, opened 2022-02-01
fix:added google font

cmcWebCode40:
- Imported google font directly from css
#133


nick-verida (approved):
I believe this should fix the issue with the font not loading after Parcel users correctly process fonts.

nick-verida:
Needs @tahpot to approve before merging

tahpot:
@cmcWebCode40 @nick-verida

We can't load fonts directly from Google as it's a GDPR risk.

See https://thehackernews.com/2022/01/german-court-rules-websites-embedding.html

We need to continue embedding the font in the library.

nick-verida (changes requested):
As mentioned by @tahpot we need to use the local URL.

I think maybe just declaring it as a module _might_ be sufficient?

cmcWebCode40:
@tahpot  noted, just knowing about this thanks!

cmcWebCode40:
@nick-verida, 
I think this PR can be closed because we stick to the previous implementation.

Issue #142, opened 2022-02-10
Add test cases for verifable credentials


nick-verida:
Test cases should cover at least:

Valid/invalid/corrupted JWT signature
Expiry of credential

Other edge cases in the Verifiable Credentials document

nick-verida:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Verida-Engineering-61ef8aa39c1cb9001f7cb0ca/issues/verida/verida-js/142?planning-poker) with ZenHub @cmcWebCode40 @tahpot

PR #143, opened 2022-02-10
feat:added test VC test cases

cmcWebCode40:
## Tests Cases
 
- Verify a JWT signature is valid
- When  a JWT signature is invalid
- Retrieve Credential data from URI using a different account

### Refactor 
- Separated create-credential and share-credential tests files
- Created a `config.ts` file to export all reusable variables and function
#142


packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    describe('Credential Units', function () {
+        this.timeout(100000);
+        let encryptedData;
+
+
+        let appContext;
+        let shareCredential;
+        let credential;
+
+        it('Login in user and create app context', async function () {
```
tahpot, line 19: This is not something that needs to be tested, as it's not credential related.

You can however put this in a "beforeEach()" so it initializes a new credential instance before each test?

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+            shareCredential = new SharingCredential(appContext);
+
+            credential = new Credentials(appContext);
+
+
+            assert.equal(appContext.getContextName(), config.VERIDA_CONTEXT_NAME);
+        });
+        it('Verify Credential JWT was created correctly', async function () {
+
+            const item: any = await credential.createCredentialJWT(config.CREDENTAIL_DATA);
```
tahpot, line 34: `CREDENTAIL_DATA` - typo

packages/verifiable-credentials/test/config.ts
```diff
@@ -0,0 +1,50 @@
+import { AutoAccount } from '@verida/account-node';
+import { EnvironmentType, Network } from '@verida/client-ts';
+
+export const config = {
+    PRIVATE_KEY_1: "0x5dd84b6d500bcbc018cbc71b0407d694095755d91af42bd3442b2dfc96b1e0af",
+    PRIVATE_KEY_2: "0x80d3b996ec98a91536efdffbae40f5eaaf117765a587483c69195c9460165c37",
+    DID_1: 'did:vda:0xB3729982A2585544FD72c99CF3773a9c6baBD55c',
+    ADDRESS_1: '0xB3729982A2585544FD72c99CF3773a9c6baBD55c',
+    CREDENTAIL_DATA: {
```
tahpot, line 9: This is not valid credential data. `alumniOf` is not a valid attribute of that credential.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+            const testJwt = {
+                didJwtVc: item.didJwtVc
+            }
+
+            assert.deepEqual(item, testJwt, 'Expected credential didJWT was generated');
```
tahpot, line 40: This tests a credential was created, but needs more tests to ensure it was generated correctly.

You should:

- Decode the credential and compare the data against the original credential data
- Compare the other credential data with the expected values (ie: issuer, timestamp etc.)

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        });
+        it('Verify a JWT signature is valid', async function () {
+
+            const item = await credential.createCredentialJWT(config.CREDENTAIL_DATA);
+
+            const data = await shareCredential.issueEncryptedCredential(item);
```
tahpot, line 46: This is testing sharing a credential. There is no need to do this, you just need to test the created credential.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+            encryptedData = data
+
+            const jwt = await Utils.fetchVeridaUri(data.uri, appContext);
+
+            const verifiedCredential: any = await credential.verifyCredential(jwt);
+
+            assert.equal(verifiedCredential.jwt, jwt, 'JWT signature is valid');
```
tahpot, line 54: This does not test the signature is valid.

The code should look at the signature and use our encryption library (and perhaps DID client library?) to ensure the signature generated is the expected value based on who signed it.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        });
+
+        it('Retrieve Credential data from URI using a different account', async function () {
```
tahpot, line 57: This is testing the ability to share credentials via URI, so it should be in the other `share` credential test file.

packages/verifiable-credentials/test/config.ts
tahpot: In fact, you should create a test for this that ensures any data sent to `createCredential()` that isn't part of the schema, isn't included in the actual credential.

packages/verifiable-credentials/test/create-credentials.test.ts
tahpot: Actually, you probably need a function that tests everything about a credential is valid, then re-use that across many of the unit tests.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+            const jwt = await Utils.fetchVeridaUri(encryptedData.uri, context);
+
+
+            const verifiedCredential: any = await credential.verifyCredential(jwt);
+
+            assert.equal(verifiedCredential.jwt, jwt, 'JWT signature is valid');
+        });
+
+        it('When  a JWT signature is invalid', async function () {
```
tahpot, line 70: I'm not sure what the purpose of this test is?

tahpot (changes requested):
See my comments below. Let me know if you have any questions.

packages/verifiable-credentials/test/create-credentials.test.ts
cmcWebCode40: @taphot , i'm not sure how to do this, is there an example test for this one  ?

cmcWebCode40:
@nick-verida this issue #145  was fixed in this commit [542d9bf](https://github.com/verida/verida-js/pull/143/commits/542d9bf66e9b82f78b9b8c6cede2f26dcbc3a79b)

packages/client-ts/src/client.ts
```diff
@@ -169,7 +175,7 @@ class Client {
    * This provides the public details about the database, storage and messaging endpoints stored on Ceramic / IDX for the requested `did`.
```
nick-verida, line 175: Correct comment to remove reference to Ceramic/IDX

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -128,29 +144,50 @@ export default class Credentials {
...
+		// Ensure a credential schema has been specified
+		if (!data.schema) {
+			throw new Error('No schema specified')
+		}
+
+		// Ensure data matches specified schema
+		const schema = await this.context.getClient().getSchema(data.schema)
+		const isValid = await schema.validate(data);
+
+		// @todo: Check the schema is a "credential" schema type?
```
tahpot, line 157: @cmcWebCode40 Can you add an issue for this?

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -128,29 +144,50 @@ export default class Credentials {
...
 			},
 			credentialSchema: {
 				id: data.schema,
 				type: 'JsonSchemaValidator2018',
 			},
 		};
-		const didJwtVc = await this.createVerifiableCredential(credential, issuer);
+
+		if (expirationDate) {
+			// @todo: verify expiration date is a valid date string
```
tahpot, line 186: @cmcWebCode40 Add a backlog issue for this as well

packages/verifiable-credentials/test/config.ts
```diff
@@ -0,0 +1,58 @@
...
+        schema: 'https://common.schemas.verida.io/social/contact/v0.1.0/schema.json'
+    },
+    INVALID_CREDENTIAL_DATA: {
+        email: 'me',
+        schema: 'https://common.schemas.verida.io/social/contact/v0.1.0/schema.json'
+    },
+    VERIDA_CONTEXT_NAME: "Verida: Credentials",
+    VERIDA_EXPECTED_DATABASE: 'credential_public_encrypted',
+    VERIDA_TESTNET_DEFAULT_SERVER: "https://db.testnet.verida.io:5002/",
+    VALID_VERIDA_URI: `verida://did:vda:0xB3729982A2585544FD72c99CF3773a9c6baBD55c/0x42406d644c8e884be0f76ee05539cfba03eda65baa2bda33ecd364ce4d18047c/credential_public_encrypted/4f54db10-8b92-11ec-b09c-bf8dd364cfc7?key=9dc07aae40b856cecb7c257b73c7929532284f5be1c153916da754cf841bdb69`,
```
tahpot, line 24: @cmcWebCode40  Are these URI's and INVALID_VC_JWT still required?

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -128,29 +145,50 @@ export default class Credentials {
...
```
nick-verida, line 186: The DID JWT VC library (called by `createVerifiableCredential`) actually does check the format of this.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        it('Ensure valid expiration date is respected', async () => {
+            // Set an expiry date to the future
+            const expirationDate = '2060-02-14T04:27:05.467Z'
+
+            const jwt: any = await credential.createCredentialJWT(config.SUBJECT_DID, config.CREDENTIAL_DATA, expirationDate);
+
+            const decodedCredential = await credential.verifyCredential(jwt.didJwtVc)
+            assert.ok(decodedCredential, 'Credential is valid')
+        });
+        /*if ('Handles invalid DID JWT', async () => {
```
tahpot, line 91: @cmcWebCode40 Remove

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
+// https://nodejs.org/api/assert.html
```
tahpot, line 1: @cmcWebCode40 Fix line spacing

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+            const decodedCredential = await credential.verifyCredential(jwt)
+
+            // Obtain the payload, that contains the verifiable credential (.vc)
+            const payload = decodedCredential.payload
+
+            const vc = payload.vc
+
+            assert.deepEqual(vc.credentialSubject, config.CREDENTIAL_DATA, 'Issuer matches expected DID');
+
+            // @todo: Fetch the credential via data.uri, decode the credential and then ensure the vc.credentialSubject matches config.CREDENTIAL_DATA
```
tahpot, line 61: @cmcWebCode40 Comment can be removed, implemented above

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        });
+        it('Retrieve Credential data from URI using a different account', async function () {
+            // @todo: Fetch the credential via data.uri, decode the credential and then ensure the vc.credentialSubject matches config.CREDENTIAL_DATA
```
tahpot, line 64: @cmcWebCode40 No longer a `todo`

tahpot (changes requested):
Added some comments on minor cleanup tasks.

packages/verifiable-credentials/src/credentials.ts
nick-verida: The DID JWT VC library (called by `createVerifiableCredential`) actually does check the format of this.

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -128,29 +144,50 @@ export default class Credentials {
...
+		const vcPayload: any = {
 			'@context': [
 				'https://www.w3.org/2018/credentials/v1',
 				'https://www.w3.org/2018/credentials/examples/v1',
 			],
-			id: '',
+			sub: subjectId,
 			type: ['VerifiableCredential'],
 			issuer: did,
 			issuanceDate: new Date().toISOString(),
```
nick-verida, line 175: Is this correct? 

In some cases the `issuanceDate` could be set externally - eg it is valid from the date payment is received, which could be prior to the VC actually being issued here. This could be important in cases where things like fines are being issued based on the date the credential is paid for.

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -88,9 +88,25 @@ export default class Credentials {
...
+	async verifyCredential(vcJwt: string): Promise<any> {
 		const resolver = this.getResolver();
-		return verifyCredential(vcJwt, resolver);
+		const decodedCredential = await verifyCredential(vcJwt, resolver);
+
+		if (decodedCredential) {
+			const payload = decodedCredential.payload
+			const vc = payload.vc
+
+			if (vc.expirationDate) {
```
nick-verida, line 99: Explain why this isn't included in the schema (see https://discord.com/channels/664594912460668960/929954765021011978/942952420433080330)

packages/verifiable-credentials/test/config.ts
cmcWebCode40: I have removed all expect for INVALID_VERIDA_URI used to test `When Verida URI data does not exist in the database`

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -88,9 +88,25 @@ export default class Credentials {
...
+				// Ensure credential hasn't expired
+				const now = new Date().toISOString()
```
nick-verida, line 101: Method should provide an optional parameter "currentDateTime" that is used instead of new Date().toISOString() if set (to allow the client to migrate cases where the datetime is incorrect on the local computer)

cmcWebCode40:
@tahpot , @nick-verida 

i have made all requested changes

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -88,9 +88,25 @@ export default class Credentials {
...
+				if (vc.expirationDate < now) {
```
nick-verida, line 102: This comparison is potentially incorrect. Both `.toISOString()` and the `XMLSCHEMA11-2` can include non-UTC timezone offsets, and this is just doing a string comparison. 

A case like this can give the incorrect result:

"2019-08-29T04:00:00+00:00:00" > "2019-08-29T04:00:00+02:00:00"

Not sure what the best solution is here - maybe convert to UTC? 

@tahpot may have suggestions.

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -128,29 +144,50 @@ export default class Credentials {
...
+
 		const issuer = await this.createIssuer();
 		const account = this.context.getAccount();
 		const did = await account.did();
 
-		const credential = {
+		const vcPayload: any = {
 			'@context': [
 				'https://www.w3.org/2018/credentials/v1',
 				'https://www.w3.org/2018/credentials/examples/v1',
```
nick-verida, line 170: Should the example be included here?

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+            // Verify the data matches the schema
+            const record = vc.credentialSubject
+            const schema = await appContext.getClient().getSchema(record.schema)
+            const isValid = await schema.validate(config.CREDENTIAL_DATA);
+            assert.equal(true, isValid, 'Credential subject successfully validates against the schema');
+
+            // Note: Don't need to check the signature, because the did-jwt-vc does this for us
+
+            // @todo: confirm vc.issuanceDate was within the last 10 seconds
```
nick-verida: Unsure about this - see previous comment about `issuanceDate`

nick-verida (approved):
A few changes. 

(I seem to have pressed "Approve" - I meant to request changes sorry!)

nick-verida (changes requested):
Changes as requested

packages/verifiable-credentials/test/create-credentials.test.ts
cmcWebCode40: > @todo: confirm vc.issuanceDate was within the last 10 seconds


@tahpot should I open a backlog issue for this?

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: Yes, it was included in the @context ref in the documentation

packages/verifiable-credentials/src/credentials.ts
tahpot: Agreed. Should support `issuanceDate` option.

@cmcWebCode40 Perhaps you refactor to support an `options` parameter that replaces `expirationDate`. There can be two options; `expirationDate`, `issuanceDate`. If `issuanceDate` is not defined, default to now.

packages/verifiable-credentials/src/credentials.ts
tahpot: Good point.

@cmcWebCode40 Add this to the `options` as well?

packages/verifiable-credentials/src/credentials.ts
tahpot: Ah yes. Javascript `toIsoString()` is always UTC, but the W3C spec allows `expirationDate` to be other timezones.

Moment.js is deprecated. [See a list of options here](https://momentjs.com/docs/#/-project-status/).

We need something that is as lightweight as possible. Maybe Luxon. https://moment.github.io/luxon/#/?id=luxon

@nick-verida Happy for you to make a call here on the best option.

packages/verifiable-credentials/test/create-credentials.test.ts
tahpot: @cmcWebCode40 Yes.

In light of the above, probably need two test cases. Test `issuanceDate` is within 10 seconds, if no `issuanceDate` provided. Otherwise do a test with a user supplied `issuanceDate`.

packages/verifiable-credentials/src/credentials.ts
tahpot: If that's the case, then remove our comment and no action required.

packages/verifiable-credentials/src/credentials.ts
nick-verida: @cmcWebCode40 The Luxon library linked above supports [ISO8601](https://moment.github.io/luxon/#/parsing?id=iso-8601) parsing and this is compatible with `XMLSCHEMA11-2` date objects. 

So use that.

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: Alright will do that

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: https://github.com/verida/verida-js/issues/147#issue-1138053083

packages/verifiable-credentials/test/create-credentials.test.ts
cmcWebCode40: Backlog issue for this change https://github.com/verida/verida-js/issues/148

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: Backlog issue for this task her 
 https://github.com/verida/verida-js/issues/146

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: I have added the `options` parameter and they are optional

packages/verifiable-credentials/src/credentials.ts
tahpot: @cmcWebCode40 Still need to remove the comment

packages/verifiable-credentials/src/credentials.ts
tahpot: @cmcWebCode40 That backlog issue isn't correct.

The required task is to use the Luxon library to do the date/time comparison.

packages/verifiable-credentials/src/credentials.ts
tahpot: I don't understand this. Leaving for @nick-verida

packages/verifiable-credentials/src/credentials.ts
tahpot: This looks good, can you also add a test to confirm the issuance date in the generated VC is the one sent in the options.

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: update the backlog task

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: @tahpot, I have added this test.

packages/verifiable-credentials/src/credentials.ts
nick-verida: @cmcWebCode40 Please replace this comment with `The DID JWT VC library (called by createVerifiableCredential) verifies the string format so we do not need a test for that`

nick-verida (changes requested):
One minor change left

tahpot (approved):
Looks good to me.

nick-verida (approved):

nick-verida:
Merged

Issue #152, opened 2022-02-21
[verifiable-credentials]: modify the URI output


cmcWebCode40:
The verifiable-credential output should provide two uri's:

veridaUri (the current uri in `data.uri`)
publicUri (the new uri `https://scan.verida.io/credential?uri={base64encode($data.uri})})`
The mobile app can then use publicUri.

PR #153, opened 2022-02-21
feat: modified uri output

cmcWebCode40:
- encoded URI's to base64
- modified test to base64  encode and decode URI's

#152


cmcWebCode40:
@tahpot, 

i have been able to fix all merge conflicts and also ensured all tests passed.

packages/verifiable-credentials/src/sharing-credential.ts
```diff
@@ -102,17 +102,20 @@ export default class SharingCredential {
...
+		if (!result) {
+			throw new Error('unable to save jwt item to db')
 		}
+		const uri = Utils.buildVeridaUri(did, contextName, dbName, result.id, params) as any
+
+		return {
+			item: item,
+			result: result,
+			did: did,
+			veridaUri: encodeBase64(uri),
```
tahpot, line 116: Don't base64 encode `veridaUrl`

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -37,11 +37,14 @@ describe('Share Credential tests', function () {
                 data.result.id
             );
 
-            const uriWithoutKey = data.uri.substring(0, expectedUri.length);
+            const base64decodedURI = base64Encoded(data.veridaUri)
```
tahpot, line 40: `base64decodedURI` should be `base64encodedURI`

packages/verifiable-credentials/test/config.ts
```diff
@@ -54,3 +73,8 @@ export const connect = async (privateKey: string, customContexName?: string): Pr
     });
     return context;
 };
+
+
+export const base64Encoded = (data: string) => {
```
tahpot, line 78: Should be `base64encode()`

tahpot (changes requested):
@cmcWebCode40 See my review comments.

@pkhien95 You seem to have added back a previous test that is now redundant. Can you please remove that code. If anything you should put it in a separate branch as this PR is now blocked by that extra code.

packages/verifiable-credentials/test/share-credentials.test.ts
cmcWebCode40: since , I'm not i won't base64encode the `veridaUrl` i don't think i will need to use this `base64Encoded` anymore

packages/verifiable-credentials/test/config.ts
cmcWebCode40: This is a mistake from me it was supposed to be `base64dencoded()` because i initially encoded the `veridaUrl`, but i think its not need anymore

cmcWebCode40:
@tahpot, 

the errors I got from the following tests show that the  `verifiableCredential.credentialSubject.didJwtVc` value is different from the `config.CREDENTIAL_DATA.didJwtVc` property value that's the cause why those tests are failing  

These are the test cases failing 
- Verify Credential JWT was created correctly
-  Issue an encrypted credential
- Retrieve Credential data from URI using a different account

tahpot (changes requested):
See comments for suggested changes.

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -64,8 +68,12 @@ describe('Share Credential tests', function () {
             const decodedPresentation = await credential.verifyPresentation(jwt)
 
             // Retrieve the verifiable credential within the presentation
+
             const verifiableCredential = decodedPresentation.verifiablePresentation.verifiableCredential[0]
 
+            deleteDidJwtProperty()
+
+            delete verifiableCredential.credentialSubject['didJwtVc'];
```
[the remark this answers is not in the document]
cmcWebCode40: removed

cmcWebCode40:
@tahpot  
i have made requested changes

pkhien95:
@tahpot @cmcWebCode40 I think the current test data in config.ts is not the actual data that credentials-issuer-demo and vault-mobile are using. That's the reason why unit tests are running smoothly but we still have blockers in credentials-issuer-demo and vault-mobile. Can we try to use the real payload which the apps are using, not "social/contact/" type of schema?

tahpot:
> @tahpot @cmcWebCode40 I think the current test data in config.ts is not the actual data that credentials-issuer-demo and vault-mobile are using. That's the reason why unit tests are running smoothly but we still have blockers in credentials-issuer-demo and vault-mobile. Can we try to use the real payload which the apps are using, not "social/contact/" type of schema?

@pkhien95 Can you create a new issue and branch with a unit test using a payload that isn't working?

You will need to include exact steps to replicate your payload from the `credentials-issuer-demo` and make sure that project has the latest `verida-js` library.

PR #161, opened 2022-02-28
changes to support move to Docusaurus for API Docs

nick-verida:
These are changes to support moving the [API docs to Docusaurus](
https://github.com/verida/documentation/issues/46).

The most significant is making sure we build sourcemaps for all modules. This doesn't add significant size but lets the `typedoc` tool work straight out-of-the-box, which is significantly easier than trying to get JSDoc to work properly against Typescript and to generate Markdown docs for Docusaurus.

Currently the docs are manually synced from this project to the `documentation` project. Automation will follow. 

See https://developers.verida.io/docs/api/verida-js/modules for how it looks. 

I anticipate writing a fair bit of documentation against the code now this is working, so if there are any significant errors in docs right now I suggest approving this PR and raising new issues will be the most useful way forward.


packages/client-ts/src/client.ts
```diff
@@ -198,7 +198,7 @@ class Client {
    *
    * @param did
    * @param contextName
-   * @returns <Profile | undefined>
+   * @returns `<Profile | undefined>`
```
nick-verida, line 201: We need the `backticks` so the Docusaurus MDX parser doesn't complain.

tahpot (approved):
This is awesome!

Issue #146, opened 2022-02-15
[verifiable-credentials]: Check the schema is a "credential" schema type


cmcWebCode40:
Check if the schema is a [credential](https://common.schemas.verida.io/credential/base/v0.1.0/schema.json) schema type

cmcWebCode40:
@tahpot, please I would want to understand exactly the task I'm implementing, when the `createCredentialJWT()` is called are my validating to check if the schema data must be a `credential` schema type ?

tahpot:
@cmcWebCode40 For now check if there is a `didJwtVc` required field in the schema.

cmcWebCode40:
@tahpot, 

```
const isValid = await schema.validate(data);
```

I think the above example (already implemented) validates the data, which means if it a credential data and it doesn't have the `didJwt` property it returns false. is that correct?

maybe this solves the issue?

Issue #148, opened 2022-02-15
[verifiable-credentials]: Test for VC issuanceDate


cmcWebCode40:
### Create two tests 

- [ ] Test issuanceDate is within 10 seconds .
- [ ]  If no issuanceDate provided. Otherwise do a test with a user-supplied issuanceDate

cmcWebCode40:
@tahpot,

Do I still need to check that the `issuanceDate` provided must be within 10secs creating Vc, what if a different date was provided and it's not close to the current date, how do we validate that?

cc @nick-verida

tahpot:
@cmcWebCode40 Only test within 10 seconds, if the `issuanceDate` wasn't supplied as a parameter.

Issue #147, opened 2022-02-15
[verifiable-credentials]: use the Luxon library to do the date/time comparison.


cmcWebCode40:
The Luxon library linked above supports [ISO8601](https://moment.github.io/luxon/#/parsing?id=iso-8601) parsing and this is compatible with XMLSCHEMA11-2 date objects.

See comments :

https://github.com/verida/verida-js/pull/143#discussion_r806374525
https://github.com/verida/verida-js/pull/143#discussion_r806386387

cmcWebCode40:
@tahpot , @nick-verida 

the luxon library size is 3.72 MB which can impact our VC lib size.
what are your thoughts?

tahpot:
@cmcWebCode40 Way too big... this is a simple comparison so need something much smaller.

Do you want to look at the other library options linked to in the original ticket?

cmcWebCode40:
@tahpot, 
sure, I will take a look at the other options.

PR #162, opened 2022-02-28
Feature/147 date comparison

cmcWebCode40:
## Tasks

 - Schema data validation
 - Use [dayjs](https://day.js.org/docs/en/parse/utc) lib for date comparison (convert to UTC time)
 - Test case to ensure issuanceDate generated in VC is within 10secs
 
 Resolves issue #146  #148 #147


packages/verifiable-credentials/src/credentials.ts
```diff
@@ -169,8 +175,14 @@ export default class Credentials {
 		// Ensure data matches specified schema
 		const schema = await this.context.getClient().getSchema(data.schema)
 		const isValid = await schema.validate(data);
+		const schemaJson = await schema.getSpecification();
+		const databaseName = schemaJson['database']['name']
 
-		// @todo: Check the schema is a "credential" schema type?
+		if (schemaJson && databaseName === 'credential') {
+			if (!data.didJwtVc) {
```
tahpot, line 182: @cmcWebCode40 You need to check that `schemaJson` has the `didJwtVc` attribute, not that the attribute is set in the credential data.

You are checking the schema has the property, not that the data matches the schema.

See the comment you deleted:

`@todo: Check the schema is a "credential" schema type?`

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -204,15 +216,24 @@ export default class Credentials {
 		}
 
 		if (options && options.issuanceDate) {
+			// validate ISO UTC date format 
+			const regex = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/
```
tahpot, line 220: Instead of doing this. Can you use the `dayjs` library to convert the string to a UTC time and then ensure it always generates the correct string format?

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -83,6 +85,14 @@ describe('Credential tests', function () {
 
             assert.deepEqual(vc.issuanceDate, issuanceDate, 'issuanceDate options matches generated VC date ');
         });
+        it('Ensure issuanceDate generated in VC is within 10secs', async () => {
+            const jwt: any = await credential.createCredentialJWT(config.SUBJECT_DID, config.CREDENTIAL_DATA);
+            const decodedCredential = await credential.verifyCredential(jwt.didJwtVc)
+            const payload = decodedCredential.payload
+            const vc = payload.vc
+
+            assert.ok(vc.issuanceDate, 'issuanceDate is within 10secs after creating ');
```
tahpot, line 94: This isn't testing issuanceDate is created within 10 seconds of creation date....

packages/verifiable-credentials/src/credentials.ts
tahpot: This also needs a unite test.

packages/verifiable-credentials/src/credentials.ts
tahpot: This also needs a unit test.

tahpot (changes requested):
See comments below.

Need to address some of the logic and make sure all new code has appropriate unit tests.

packages/verifiable-credentials/test/create-credentials.test.ts
cmcWebCode40: @tahpot,
Please, I think I would need help here, how do I implement this, how do I compare the time, its unclear to me .

tahpot: You need to compare the `issuanceDate` is within 10 seconds of the current time.

The idea is that the verifiable credentials library automatically set the `issuanceDate` and we want to test it was created at the current time. It doesn't need to be 10 seconds, it could be 5.

cmcWebCode40: This would happen only when the `issuanceDate` is not provided right?

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: modified this logic and added a test case

cmcWebCode40: implemented

packages/verifiable-credentials/test/create-credentials.test.ts
tahpot: Yes, correct.

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -206,21 +218,38 @@ export default class Credentials {
...
+
+			vcPayload.issuanceDate = dateFormat.$d
 		} else {
-			vcPayload.issuanceDate = new Date().toISOString()
-		}
+			const currentTime = dayjs(new Date().toISOString()).utc(true).second();
+			const createdVcTime = dayjs(vcPayload.issuanceDate).utc(true).second()
+			const MAX_DURATION_IN_SECONDS = 5
 
+			if ((currentTime - createdVcTime) > MAX_DURATION_IN_SECONDS) {
```
tahpot, line 245: @cmcWebCode40 This is incorrect

You need a unit test that checks the `issuanceDate` is created within 5 seconds of the current time.

We don't want to restrict credentials to be created within a particular time frame.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -81,16 +101,34 @@ describe('Credential tests', function () {
...
+            const decodedCredential = await credential.verifyCredential(jwt.didJwtVc)
+            const payload = decodedCredential.payload
+            const vc = payload.vc
+
+            const currentTime = dayjs(new Date().toISOString()).utc(true).second();
+            const createdVcTime = dayjs(vc.issuanceDate).utc(true).second()
+
+            const issuanceDate = currentTime - createdVcTime;
+
+            assert.ok(issuanceDate < 5, 'issuanceDate is within 10secs after creating ');
```
tahpot, line 121: Should be `10` not `5`

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -177,7 +185,11 @@ export default class Credentials {
 		dataClone['didJwtVc'] = 'ABC'
 		const isValid = await schema.validate(dataClone);
 
-		// @todo: Check the schema is a "credential" schema type?
+		if (schemaJson && databaseName === 'credential') {
+			if (!schemaJson.properties.didJwtVc) {
+				throw new Error('please provide the didJwtVc property for credential schema type')
```
tahpot, line 190: Make error say `Schema is not a valid credential schema`

tahpot (changes requested):
See comments.

packages/verifiable-credentials/src/credentials.ts
cmcWebCode40: @tahpot, 

here is the test case i wrote for this https://github.com/verida/verida-js/blob/1e99a864f4d141afe928954157896921db65f623/packages/verifiable-credentials/test/create-credentials.test.ts#L110

is this test case incorrect ?

tahpot: Yes, that test case is correct (apart from my comment below)

cmcWebCode40: I think I'm a little bit confused on how exactly how to implement this feature

PR #163, opened 2022-03-01
Add real credentials data in test case

pkhien95:
In https://github.com/verida/demo-credential-issuer/issues/39, we need to create the `didJwtVc` property based on the data user filled in the form.
Currently when the app calls `credential.createCredentialJWT()` (full example: https://github.com/verida/demo-credential-issuer/blob/96629599c9fad62c5d539435871b24323e00ee99/src/helpers/VeridaClient.helpers.ts#L21), it will throw this error:
```
credentials.js?757d:246 Uncaught (in promise) Error: Data does not match specified schema
    at Credentials.eval (credentials.js?757d:246:1)
    at step (credentials.js?757d:44:1)
    at Object.eval [as next] (credentials.js?757d:25:1)
    at fulfilled (credentials.js?757d:16:1)
```
The current test case in this PR is for reproducing and debugging this issue.

### DIFF
--- a/packages/verifiable-credentials/src/credentials.ts
+++ b/packages/verifiable-credentials/src/credentials.ts
@@ -168,11 +168,19 @@ export default class Credentials {
 
 		// Ensure data matches specified schema
 		const schema = await this.context.getClient().getSchema(data.schema)
-		const isValid = await schema.validate(data);
+
+		// Before validating, we need to ensure there is a `didJwtVc` attribute on the data
+		// `didJwtVc` is a required field, but will only be set upon completion of this
+		// creation process.
+		// @see https://github.com/verida/verida-js/pull/163
+		const dataClone = Object.assign({}, data);
+		dataClone['didJwtVc'] = 'ABC'
+		const isValid = await schema.validate(dataClone);
 
 		// @todo: Check the schema is a "credential" schema type?
 
 		if (!isValid) {
+			this.errors = schema.errors
 			throw new Error('Data does not match specified schema')
 		}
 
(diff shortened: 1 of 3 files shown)


cmcWebCode40:
@tahpot,

concerning this issue @pkhien95  raised, I have noticed that there is one more use case that has not been added.

from the  `credential-issuer-demo` here is an example data 

```
const    credentialData = {
        "name": "Your Dentist Credential",
        "firstName": "Hien ",
        "lastName": "Verida",
        "regNumber": "KSSKJSKWPO2202",
        "healthType": "Dentist",
        "regExpDate": "2022-02-02",
        "schema": "https://verida.github.io/demo-credential-issuer/mapay/v0.1.0/schema.json",
        "testTimestamp": "2022-02-17T05:54:48.834Z",
        "summary": "Credential issued at Thu Feb 17 2022",
    }
``` 

so when I call the `credential.createCredentialJWT(subjectDID,credentialData )` method in the demo app  the code throws the above validation error [where it validates from the lib](https://github.com/verida/verida-js/blob/f25cf82412f7c2cca5b85841caedba4151784680/packages/verifiable-credentials/src/credentials.ts#L187) @pkhien95 stated because the `credentialData`  does not contain the `didJwtVC` property which is required in the specified schema `https://verida.github.io/demo-credential-issuer/mapay/v0.1.0/schema.json`

tahpot:
FYI: @pkhien95 The branch name `add_real_credential_data_in_test_case` doesn't match our coding standards. There should be an issue created for this problem and the branch referencing the issue, ie: `fix/163-verifiable-credential-validationt-error`.

In the future, pleas ensure the name describes the problem. `Add real credentials data in test case` specifies what you are doing, but not what the problem is.

pkhien95:
> FYI: @pkhien95 The branch name `add_real_credential_data_in_test_case` doesn't match our coding standards. There should be an issue created for this problem and the branch referencing the issue, ie: `fix/163-verifiable-credential-validationt-error`.
> 
> In the future, pleas ensure the name describes the problem. `Add real credentials data in test case` specifies what you are doing, but not what the problem is.

Sorry I was not sure what is the root cause behind this so the branch is just for debugging purpose until someone finds out what exactly is the issue. Now we have found out that the issue is about "validation error", I created [#164](https://github.com/verida/verida-js/pull/165) and migrated this PR to https://github.com/verida/verida-js/pull/165

PR #175, opened 2022-03-25
Remove logoUrl in VaultAccountConfig

pkhien95:
This is a support PR to resolve https://github.com/verida/vault-mobile/issues/394


nick-verida:
The current syntax is correct. See https://github.com/verida/verida-js/pull/135

PR #187, opened 2022-04-25
Feature/185 update create credential jwt

cmcWebCode40:
- [x] Added veridaContextName in vc object

Fixes #187


tahpot (approved):
Thanks @cmcWebCode40 -- looks good :)

cmcWebCode40:
@tahpot, 

The new change I made was to modify the` createCredentialJWT()` method so it accepts the `veridaContextName` used in opening a  public profile.

## The Issue 

- We tried sending creating a VC from the `demo-credential`  app, but the logged-in `issuer`'s profile `contextName` from the demo-app, is `Verida: Vault` meanwhile the library tries to use the issuer app `contextName`   which is `Verida: Verifiable Credential Demo` different from the profile contextName  `Verida: Vault`, this caused an error in the account-explorer-demo because it was unable to look up the profile with the app `contextName` (Verida: Verifiable Credential ).

But this works for the `credential-mgt-platform` demo because the same app contextName was for used setting a public profile.

## Solution
So we have to pass the  profile `contextName` as an argument in the` createCredentialJWT()` which solves this problem

what are your thoughts on this approach?

tahpot:
@cmcWebCode40 

No, this change is not correct.

The `account-explorer-demo` needs to lookup the the profile with the app context name (Verida: Verifiable Credential). If not found, it needs to lookup `Verida: Vault` as a fallback.

This is the logic used by the Vault.

cmcWebCode40:
> so we may need to consider a bit of tooling in the SDK to make this easier?
 using the Verida: Vault as a fallback sounds good I will have to test again and  @andy-verida will test as well in the vault.

andy-verida:
@tahpot, @cmcWebCode40 and I tested together and confirm the fallback approach to the default context `Verida: Vault` works. the code looks sth like this
```
const MAX_PROFILE_RETRY = 2
export async function getPublicProfile(
  did: string,
  contextName = VERIDA_CONTEXT_NAME
) {
  let ctxName = contextName
  const retryNumber = ctxName === VERIDA_CONTEXT_NAME ? 1 : MAX_PROFILE_RETRY
  const client = AccountManager.getInstance().getClient()
  for (let t = 1; t <= retryNumber; t++) {
    if (t === MAX_PROFILE_RETRY) {
      ctxName = VERIDA_CONTEXT_NAME
    }

    try {
      const publicProfile = await client?.openPublicProfile(
        did,
        ctxName,
        'basicProfile'
      )
      const name = await publicProfile?.get('name')
      const avatar = await publicProfile?.get('avatar')

      return {
        name: name || 'Unknown',
        avatar: avatar || DefaultAvatar,
      }
    } catch (error: any) {
      Sentry.captureException(error)
    }
  }
}
```

tahpot:
> Though I prefer code to be implemented at the app level, but not sure should we make it into the `openPublicProfile` function inside client lib so we can avoid repeating code in many apps?

I think this will be a common issue, so yes lets do that.

What if there is an extra fourth parameter to `openPublicProfile()` called `fallbackContext = "Verida: Vault"`?

It will default to "Verida: Vault", but could be set to nothing if the app doesn't want the fallback.

@andy-verida Feel free to create an issue and PR with that change if you agree.

andy-verida:
> What if there is an extra fourth parameter to `openPublicProfile()` called `fallbackContext = "Verida: Vault"`?

@tahpot looks good, I created a new issue here https://github.com/verida/verida-js/issues/194

tahpot (changes requested):
This can not be approved as it adds `issuerName` functionality that is not necessary.

packages/verifiable-credentials/test/create-credentials.test.ts
```diff
@@ -20,17 +20,23 @@ describe('Credential tests', function () {
 
         beforeEach(async function () {
 
-            appContext = await connectAccount(config.PRIVATE_KEY_1, config.VERIDA_CONTEXT_NAME, EnvironmentType.TESTNET);
+            appContext = await connectAccount(config.PRIVATE_KEY_1, config.VERIDA_CONTEXT_NAME, EnvironmentType.TESTNET, config.ISSUER_NAME);
```
tahpot, line 23: issuer name

packages/verifiable-credentials/src/utils.ts
```diff
@@ -11,7 +11,7 @@ import config from './config';
  * @returns  an application {context} of the connected account
  */
 
-export async function connectAccount(privateKey: string, contextName: string, environment: EnvironmentType): Promise<Context> {
+export async function connectAccount(privateKey: string, contextName: string, environment: EnvironmentType, issuerName: string): Promise<Context> {
```
tahpot, line 14: issuer name

packages/verifiable-credentials/test/share-credentials.test.ts
```diff
@@ -16,7 +16,7 @@ describe('Share Credential tests', function () {
         let createdUri = ''
 
         beforeEach(async function () {
-            appContext = await connectAccount(config.PRIVATE_KEY_1, config.VERIDA_CONTEXT_NAME, EnvironmentType.TESTNET);
+            appContext = await connectAccount(config.PRIVATE_KEY_1, config.VERIDA_CONTEXT_NAME, EnvironmentType.TESTNET, config.ISSUER_NAME);
```
tahpot, line 19: issuer name

tahpot (changes requested):
Tagging issuer name references

tahpot (approved):
Thanks @cmcWebCode40, all approved.

PR #193, opened 2022-05-05
Fix/190 get many return array. Closes 190

nick-verida:
getMany() currently returns undefined if CouchDB returns falsey. @tahpot suggested it should return a zero-length array in those circumstances. 

This PR implements that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ ipfs
 node_modules
 .DS_Store
 *.log
+api-docs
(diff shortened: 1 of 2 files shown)


nick-verida:
Adding @pranavburnwal as a reviewer

pranavburnwal:
Looks okay to me as at least this will be returning an empty array. Unsure of usage of the returned value but will not breath any direct code.

pranavburnwal (approved):
Does .gitignore update has anything to do with this PR?

Rest okay.

nick-verida:
> Does .gitignore update has anything to do with this PR?
> 

@pranavburnwal  It actually does! This change modified the return signature so the docs need redeployment. Testing that and I almost checked them in by mistake...

nick-verida:
Merging this

Issue #194, opened 2022-05-13
Add an extra param `fallbackContext` to `openPublicProfile` method default value to `Verida: Vault`


andy-verida:
Follow comment here: https://github.com/verida/verida-js/pull/187#issuecomment-1125491098

PR #200, opened 2022-05-18
Add an extra param `fallbackContext` to `openPublicProfile` method default value to `Verida: Vault`

andy-verida:
## Changes

- Add an extra param `fallbackContext` to `openPublicProfile` method default value to `Verida: Vault`
- Add tests

Issue: https://github.com/verida/verida-js/issues/194

### DIFF
--- a/packages/client-ts/src/client.ts
+++ b/packages/client-ts/src/client.ts
@@ -203,9 +203,29 @@ class Client {
   public async openPublicProfile(
     did: string,
     contextName: string,
-    profileName: string = "basicProfile"
+    profileName: string = "basicProfile",
+    fallbackContext: string = "Verida: Vault"
   ): Promise<Profile | undefined> {
-    const context = await this.openExternalContext(contextName, did);
+    const maxRetry = 2;
+    let context: Context | undefined = undefined;
+    let retryNumber = contextName === fallbackContext ? 1 : maxRetry;
+    let ctxName = contextName;
+
+    do {
+      try {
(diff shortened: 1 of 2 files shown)


tahpot (approved):
@andy-verida This is approved for merging.

However, see my comment to consider.

packages/client-ts/src/client.ts
```diff
@@ -203,9 +203,29 @@ class Client {
   public async openPublicProfile(
     did: string,
     contextName: string,
-    profileName: string = "basicProfile"
+    profileName: string = "basicProfile",
+    fallbackContext: string = "Verida: Vault"
   ): Promise<Profile | undefined> {
-    const context = await this.openExternalContext(contextName, did);
+    const maxRetry = 2;
```
tahpot, line 209: An alternative implementation here could avoid using a retry count.

Basically, when the error is caught (L220), if `fallbackContext` isn't empty or null, return this.openPublicProfile(...) and set `fallbackContext` to null. Next time the error is caught, it will be rethrown because there is no remaining `fallbackContext`.

andy-verida: You're right. Thanks @tahpot

andy-verida: updated, funny how I wrote so much verbose code 🥲 

@tahpot should I merge this PR right away or wait for later.

cmcWebCode40 (commented):
Looks good

PR #203, opened 2022-05-19
Fix infinite recursion cause of set `undefined` to `fallbackContext`

andy-verida:
## Changes
- Hotfix infinite recursion cause of set `undefined` to `fallbackContext`
- Add test

```
public async openPublicProfile(
    did: string,
    contextName: string,
    profileName: string = "basicProfile",
    fallbackContext: string | null = "Verida: Vault"
  ): Promise<Profile | undefined> {
    let context: Context | undefined;
    try {
      context = await this.openExternalContext(contextName, did);
    } catch (error) {
      if (fallbackContext) {
        // !!! Set `undefined` to the fourth param of function has no effect, the default value will be picked up anyway
        return await this.openPublicProfile(did, fallbackContext, profileName, undefined); 

        // Change to this
        // return await this.openPublicProfile(did, fallbackContext, profileName, null);  <= Null
      }
    }

    if (!context) {
      throw new Error(
        `Account does not have a public profile for ${contextName}`
      );
    }

    return context!.openProfile(profileName, did, false);
  }
```

### DIFF
--- a/packages/client-ts/src/client.ts
+++ b/packages/client-ts/src/client.ts
@@ -204,14 +204,14 @@ class Client {
     did: string,
     contextName: string,
     profileName: string = "basicProfile",
-    fallbackContext: string | undefined = "Verida: Vault"
+    fallbackContext: string | null = "Verida: Vault"
   ): Promise<Profile | undefined> {
     let context: Context | undefined;
     try {
       context = await this.openExternalContext(contextName, did);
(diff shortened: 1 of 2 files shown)


tahpot (approved):
Thanks @andy-verida, I also assumed `undefined` and `null` would behave the same here.

Great work adding a unit test for this as well.

PR #211, opened 2022-06-15
Feature/210 decentralize did registry

### LINKED ISSUE #210: [did-client] Migrate to decentralized DID registry
Use https://github.com/verida/blockchain-vda-web3-client to connect to the verida did registry on-chain.

See a video here describing the changes: https://www.loom.com/share/a695651dc27a4a7abf5e0a9c996d3677?from_recorder=1&focus_title=1

tahpot:
Will close #210.


pranavburnwal:
In behest of #213, This PR is outdated. 

cc @ITStar10 @tahpot

tahpot:
This is updating did-client

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
-     * @returns 
+     * @param document Updated DIDDocuent
+     * @returns true if success.
      */
-    public async save(document: DIDDocument): Promise<boolean> {
-        if (!this.privateKey) {
-            throw new Error("Unable to save DIDDocument. No private key.")
+    public async saveDocument(document: DIDDocument | undefined): Promise<boolean> {
+        if (!this.didDoc) {
+            return Promise.reject("No original document")
```
tahpot, line 232: It should be possible to save a new document.

If `this.didDoc` doesn't exist, create a new empty document.

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+    private identifier: string
+
+    // Verida Wallet Info
+    private veridaPrivateKey: Uint8Array
+    private veridaWallet: VeridaWallet
+
+    // private did?: string
+    
+    // Verida Doc
+    private didDoc? : DIDDocument
```
tahpot, line 107: DID Client can not store DID documents. Its designed to be called to fetch and save did documents, but not retain any state information.

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+
+
+    // vda-did resolver
+    private didResolver: Resolver
+    private vdaDid: VdaDID
+
+    constructor(config: DIDClientConfig) {
+        this.config = config
+
+        this.identifier = config.identifier
```
tahpot, line 117: This should be calculated from the VDA private key

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+        //             id: 'did:ethr:0x61:0x599b3912A63c98dC774eF3E60282fBdf14cda748?context=0xcfbf4621af64386c92c0badd0aa3ae3877a6ea6e298dfa54aa6b1ebe00769b28&type=database',
+        //             type: 'VeridaDatabase',
+        //             serviceEndpoint: 'https://db.testnet.verida.io:5002'
+        //         }
+        //     ]
+        //   }
     }
 
+    public authenticate(privateKey: string) {
+        throw new Error('Method not implemented.')
```
tahpot, line 213: This method accepts a private key and then unlocks the ability to save documents.

It should be possible to "get" a document, without providing a private key.

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+        const web3Config = config.callType === 'gasless' ?
+            <VeridaMetaTransactionConfig>config.web3Config :
+            <VeridaSelfTransactionConfig>{
+                ...<VeridaSelfTransactionConfigPart>config.web3Config,
+                provider: config.provider,
+                rpcUrl: config.rpcUrl,
+                web3: config.web3
+            }
+
+        this.vdaDid = new VdaDID({
```
tahpot, line 142: See below, this should only happen when the user "authenticate()"s.

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+ */
+export interface DIDClient {
+
+    authenticate(privateKey: string): void;
+
+    getDid(): string | undefined;
+
+    getPublicKey(): string;
+
+    saveDocument(document: DIDDocument): Promise<boolean>;
```
tahpot, line 75: Rename to "save()" to match the previous interface to avoid breaking other protocol code

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+
+    getDocument(): Promise<DIDDocument | null>
```
tahpot, line 77: Rename to "get()" to match the previous interface to avoid breaking other protocol code

packages/did-client/src/did-client.ts
```diff
@@ -1,90 +1,308 @@
...
+}
+
+interface Controller extends VerificationMethodBase {
+    blockchainAccountId: string
+}
+
+/**
+ * Interface to DIDClient instance
+ */
+export interface DIDClient {
```
tahpot, line 67: This interface needs to match the methods the previous did-client implementation had.

packages/did-client/src/did-client.ts
ITStar10: Removed.

packages/did-client/src/did-client.ts
ITStar10: Created an empty document when original document is not existed.

packages/did-client/src/did-client.ts
ITStar10: Updated.

packages/did-client/src/did-client.ts
ITStar10: Once privateKey not provided in constructor, there should be an identifier in config. If not we can't create a vda-did-resolver and consequently not able to get document.

packages/did-client/src/did-client.ts
ITStar10: done

packages/did-client/src/did-client.ts
ITStar10: done

packages/did-client/src/did-client.ts
ITStar10: Will remove interface and export class directly.
This interface was designed to load did document asynchronously in the did-document creation.
Not necessary now.

ITStar10 (commented):
Should remove unnecessary test files in vda-did-xx libraries.

packages/did-client/src/did-client.ts
ITStar10: There is no verida private key in constructor params now.
Verida private key will be inputed in authenticate() function.
Now in the constructor, did-client interpret the inputed identifiers and deduce DID from it. Input param for `identifier` can be an DID or an address value.

packages/did-client/src/did-client.ts
ITStar10: Implemented.

packages/did-client/src/did-client.ts
ITStar10: Done.

tahpot:
Next steps for @ITStar10 in `vda-did-resolver`:

1. Update the configuration to have `testnet` that uses Verida contract address and chainId. Remove all the ethereum networks.
2. Replace `ethr` with `vda` for all the code, so the library expects `vda` DID's.

`did-document`:

1. @tahpot to produce a broken unit test to show this issue

tahpot:
@ITStar10 I have added back the original unit tests in `did-client`.

None of them pass, but we need them to all pass. Can you please review and investigate what is needed to change so they all pass?

packages/did-document/src/did-document.ts
```diff
@@ -101,10 +103,10 @@ export default class DIDDocument {
             return !entry.id.match(contextDid)
         })
         this.doc.assertionMethod = this.doc.assertionMethod!.filter((entry: string | VerificationMethod) => {
-            return entry !== `${this.doc.id}?context=${contextHash}#sign`
+            return entry !== `${this.doc.id}?context=${contextHash}`
```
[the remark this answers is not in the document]
ITStar10: image
Please take a look for Key purpose of public keys.
As there are purpose of keys while adding by setAttribute() function, we don't need feel to set key type after context.
As you can see, there are 3 key purposes of veriKey, sigAuth, & enc.
Here I matched "#sign" to sigAuth & "#asym" to veriKey.
Do we need to add more key types?
And I am not sure whether adding more types is compatible with DID spec.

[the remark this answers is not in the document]
ITStar10: Link is https://github.com/decentralized-identity/ethr-did-resolver/blob/master/doc/did-method-spec.md

tahpot:
I have merged #231 which solves the following issues:

1. The code now expects `vda` instead of `ethr` in the did registry client library
2. Original `did-client` unit tests have been added back and now pass
3. Default did registry configuration uses testnet and previous ethereum networks have been removed

The following issues remain:

1. There are lots of commented out tests in `did-client.basic.test.ts`. Do you want to remove these or uncomment them?
2. `chainId` being inserted into `did` (this is not part of the VDA spec and will break everything)
3. Inability to handle both `asym` and `sym` public keys for an application context

packages/did-document/src/did-document.ts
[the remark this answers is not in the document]
tahpot: I have resolved this issue with this commit: https://github.com/verida/verida-js/pull/211/commits/b069bd590456a95c9134cd04f23994583fd545ed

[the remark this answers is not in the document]
tahpot: Actually, I haven't resolved this.

How can `did-document` specify the key purpose so that it is saved? I think that's the problem. This code needs to set the key purpose so that `vda-did-resolver` can save it to the blockchain.

ITStar10:
- did-client.basic.test.ts file is already removed.
- chainId is for DID-Registry-contracts deployed on the blockchains other than Polygon.
If we're aiming at Polygon only, we can remove chainId in the configuration. chainId is used to create RPCProvider.
Instead of chainId in the configuration, we can use only chainName and we should have a list of chain IDs of supported chain name in our code.
- As mentioned before, please check `addContextSignKey()` and `addContextAsymKey()` functions in the did-document.ts.
- Will check storage-link.

ITStar10:
> There are lots of commented out tests in did-client.basic.test.ts. Do you want to remove these or uncomment them?
This file is already removed.

packages/did-document/src/did-document.ts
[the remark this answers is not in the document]
ITStar10: Please take a look line#93 ~ line#100 of did-client/src/helpers.ts.
You can know that how it deduce the keyPurpose from created DIDDocument.
For each item in the document.verificationMethod, if same id is contained in document.authentication the keyPurpose is 'sigAuth'. If same id is contained in document.keyAgreemtn, keyPurpose is 'enc'.

image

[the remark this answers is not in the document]
tahpot: How does this map back to the DID document when the smart contract is read?

Can we make this more explicit by using query parameters in the publicKey that specify “typ”? That logic will not always hold true in the future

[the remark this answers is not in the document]
ITStar10: Please check line#266 ~ line#270 in the vda-did-resolver/src/resolver.ts. Same logic. I created above code from the vda-did-resolver. They're matched.

image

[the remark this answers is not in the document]
ITStar10: Of course, we can add type in the attribute value. To do so, need to check out vda-did & vda-did-resolver again.

tahpot:
@ITStar10 I have updated the did-client tests to show how they don't work.

See [this commit](https://github.com/verida/verida-js/pull/211/commits/eebf84ec550d8e19a2c32580e9ad144bafff2c94)

I have updated the test to do a comparison on the original DID document that was to be saved and then on the DID document retrieved from the blockchain.

These are the issues:

1. `verificationMethod`: `type=asym` and `type=sign` values are not being saved. `context` not being saved.
2. `keyAgreement`: `type=asym` is not being saved. `context` not being saved.
3. `assertionMethod`: `type=sign` is not being saved. `context` not being saved.

PR #224, opened 2022-08-10
EncryptionUtils sign() function support hex values

### LINKED ISSUE #223: [encryption-utils] Upgrade the signData() function to sign Hex values properly
What doesnt work:
https://github.com/verida/verida-js/blob/main/packages/encryption-utils/src/index.ts#L109

to sign data passed in as Hexadecimal values(eg hashes)

What works:
https://github.com/pubkey/eth-crypto/blob/master/src/sign.js

Reference comment from NONA:
https://github.com/verida/private-data-bridge/blob/main/packages/ethereum/tests/data-signing-options.test.js#L67

ITStar10:
Updated:
	hash() function

Tested:
	Tested for hex values & arrays

Closes #223

### DIFF
--- a/packages/encryption-utils/src/index.ts
+++ b/packages/encryption-utils/src/index.ts
@@ -7,6 +7,7 @@ import {
 } from "tweetnacl-util";
 import { ethers, utils } from 'ethers'
 import util = require("tweetnacl-util");
+import { isHexString } from "ethers/lib/utils";
 
 const newSymNonce = () => randomBytes(secretbox.nonceLength);
 const newAsymNonce = () => randomBytes(box.nonceLength);
@@ -151,7 +152,12 @@ export default class EncryptionUtils {
     }
 
     static hash(data: any) {
-        if (typeof(data) != 'string' && !isArrayLike(data)) {
+        
+        if (typeof(data) === 'string') {
+            if (!isHexString(data)) {
+                data = utils.toUtf8Bytes(data)
+            }
+        } else if (!isArrayLike(data)) {
             data = utils.toUtf8Bytes(JSON.stringify(data))
         }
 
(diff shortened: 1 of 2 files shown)


packages/encryption-utils/test/all.tests.ts
```diff
@@ -12,6 +12,11 @@ const signingKey = {
     privateKey: new Uint8Array(Buffer.from(wallet.privateKey.substr(2),'hex'))
 }
```
pranavburnwal, line 14: Please remove these console.log

pranavburnwal (changes requested):

packages/encryption-utils/test/all.tests.ts
ITStar10: It's a randomly created key.

ITStar10: Updated

packages/encryption-utils/src/index.ts
```diff
@@ -137,11 +152,16 @@ export default class EncryptionUtils {
...
 
     static hash(data: any) {
-        if (typeof(data) != 'string') {
-            data = JSON.stringify(data)
+        
+        if (typeof(data) === 'string') {
+            if (!isHexString(data)) {
+                data = utils.toUtf8Bytes(data)
+            }
+        } else if (!isArrayLike(data)) {
```
tahpot, line 160: If the data is an array, it will no longer be JSON encoded. That is a change from the previous behaviour.

Is that intentional?

packages/encryption-utils/test/all.tests.ts
```diff
@@ -73,7 +73,8 @@ describe('Encryption tests', () => {
         })
 
         it('can create and verify signatures of a string input', async () => {
-            const input = 'hello world'
+            // const input = 'hello world'
+            const input = '0x1234'
```
tahpot, line 77: Need to also check for `hello world`

tahpot (changes requested):
Added some comments.

cc @ITStar10 @pranavburnwal

packages/encryption-utils/src/index.ts
ITStar10: Not only that. If data is hex string, it will not convert it to Utf8bytes.

tahpot (approved):
This looks good.

I ran the tests locally and they all pass. Code review looks good.

PR #215, opened 2022-07-13
Move vda-did, vda-did-resolver, vda-web3-clinet

ITStar10:
Resolves #216  #217

PR #226, opened 2022-09-01
feature/218 Update DID client to support get() and set()

tahpot:
Resolves #215


tahpot:
Initial commit that adds support for `DIDDocument.compare()`. Handing over to @ITStar10 to continue from here.

PR #234, opened 2022-10-04
Feature/210 refactor

tahpot:
Fixes to:

- `did-document`
- `did-client`
- `storage-link`

Refactor of `did-client` to be simpler and use a single `DIDClient` export as per the original code.


packages/did-client/src/did-client.ts
```diff
@@ -32,104 +23,39 @@ interface VeridaSelfTransactionConfigPart  {
...
-
-    getDid(): string | undefined;
-
-    getPublicKey(): string | undefined;
-
-    save(document: DIDDocument): Promise<boolean>;
-
-    get(did?: string): Promise<DIDDocument | null>
-    
+    network: string             // `testnet` OR `mainnet`
```
ITStar10, line 26: Users can't create DIDClient with chain ID. It's now supporting for only mainnet or testnet.
Question : Verida will not support for other chains later? It will run on only Polygon network?
If we need to support another chains, we should update this configuration.

packages/did-client/src/did-client.ts
```diff
@@ -32,104 +23,39 @@ interface VeridaSelfTransactionConfigPart  {
...
+    connectMode: string         // direct OR gasless @todo: use proper enums
+    rpcUrl : string            // blockchain RPC URI to use
```
ITStar10, line 28: Before updates, users could create a did-client with one of signer, provider, or rprcUrl.
Less configuration makes it clear but users will lose diversity of creation.

packages/did-client/src/did-client.ts
tahpot: Noted.

We don't plan to support other chains at this stage. If we did, we will add them to the `vda-did-resolver` list of supported networks.

packages/did-client/src/did-client.ts
tahpot: Correct. For now, I'm fine to restrict to `rpcUrl` or `gasless`.

PR #236, opened 2022-10-05
235 - Add publicKeyHex fields to each item in VerificationMethod

ITStar10:
Update done.

Tested with test code of did-client.

Missing : Last test case is being failed.


tahpot (approved):
Looks good

PR #250, opened 2022-11-04
DID-Document serviceEndpoint type updated

ITStar10:
did-document/src/interface
	interface SecurecontextEndPoint{
		...
		endpointUri: string[] //updated from string to string[]
	}


tahpot:
No longer relevant. Fixed in `acacia` release PR.

PR #267, opened 2022-11-25
Vda-web3-client Gas configuration updated

ITStar10:
Tested
	- for each gas configuration type


tahpot (commented):
Two key changes:

1. Leave existing gas calculation code as a fallback if no gas config is provided
2. Gas limit per method is the units not the fee.

packages/vda-web3-client/src/VeridaContractBase.ts
```diff
@@ -203,16 +209,29 @@ export class VeridaContract {
...
+                    // let gasLimit = await contract.estimateGas[methodName](...params);
+                    // gasLimit = gasLimit.mul(BigNumber.from(15)).div(BigNumber.from(10)) // Multiply 1.1
+                    // console.log(gasLimit.toString())
+
+                    const gasConfig : Record<string, any> = {}
+                    // {
+                    //     gasLimit: 500000 // gasLimit,
+                    //     // gasPrice
+                    // }
+                    if (this.web3Config?.fixedGasPerMethod?.has(methodName)) {
```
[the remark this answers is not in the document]
ITStar10: They're used as a gasLimit already.
image

packages/vda-web3-client/src/VeridaContractBase.ts
```diff
@@ -206,40 +209,45 @@ export class VeridaContract {
                 if (methodType === 'view') {
                     ret = await contract.callStatic[methodName](...params)
                 } else {
-                    if (this.logPerformance) {
-                        console.log('Node endpoint: ' + (<JsonRpcProvider> contract.provider).connection.url)
-                    }
+                    let gasConfig : Record<string, any> | undefined = {}
+                    // console.log("Params : ", params)
+
+                    const paramLen = params.length
```
tahpot, line 215: Please specify a typescript type for `params`

packages/vda-web3-client/src/VeridaContractBase.ts
```diff
@@ -206,40 +209,45 @@ export class VeridaContract {
...
+                    if (params !== undefined
```
tahpot, line 216: This is very difficult to read.

Perhaps try doing something like:

const gasConfig = _.merge(params, this.methodDefaults[methodName], this.webConfig.gasConfig)

packages/vda-web3-client/test/didregistry.test.ts
```diff
@@ -402,4 +402,47 @@ describe('DidRegistry Test', () => {
       );
     });
   });
+
+  describe('Manual gas configuration test', () => {
```
tahpot, line 406: This is testing the gasConfig being specified as a param.

Can we have tests for `methodDefaults` and `web3config` settings?

tahpot (changes requested):

packages/vda-web3-client/test/didregistry.test.ts
ITStar10: Global gas configuration & `methodDefaults` is already defiend in the "test/utils.ts" file at line#86 ~ line#95

tahpot: They are defined there, but you need to tests that show the `web3config` is used if the other options aren't provided and that `methodDefaults` is used if gas config isn't specified in `params`.

tahpot (changes requested):
Looks good, just one comment remaining on the test.

packages/vda-web3-client/test/didregistry.test.ts
ITStar10: Already tested for various cases. Please check `didregistry.test.ts` file again:
line#411 : `No gas configuration` is test for default web3 config. 
line#425 : `Call wigh global gas configuration` : Used the global gas cofiguration
line#459 : `Call with method default gas configuration` : Used the methodDefaults configuration with global configuration defined
line#513 : `Call with manual gas configuration` : Used to test gas configuration in the function parameters

tahpot: Got it! I missed that. Looks great :)

tahpot (approved):

PR #268, opened 2022-11-28
Feature/192 Add supporting named accounts

### LINKED ISSUE #192: Support named verida accounts
Use the `name-registry` to lookup verida accounts with a specific name.

tahpot:
Closes #192


tahpot:
@ITStar10 I have this basically working, except when I try to run it against the smart contract deployed on Polygon Mumbai, I always get `invalid signature`. The signing code was copied directly from your working unit tests, so I don't know what the problem is. Perhaps Mumbai doesn't have the latest contract deployed or I'm using the wrong contract address?

Contract address is specified here: https://github.com/verida/verida-js/pull/268/files#diff-4c679fe299c1df8851850945c1ee983f06275b657f595ce00d6d9243e714c089R6

tahpot:
Please update the default postfix for Verida names to be `.vda` instead of `.verida`.

tahpot:
@ITStar10 Can you please add this library to standardize usernames?

https://github.com/adraffy/ens-normalize.js

PR #278, opened 2023-01-24
Create a re-usable WebUser class

tahpot:


packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+export interface WebUserProfile {
+    name?: string
+    avatarUri: string
+    country?: string
+    description?: string
+}
+
+export interface WebUserConfig {
+    accountConfig: VaultAccountConfig
+    clientConfig: ClientConfig
```
aurelticot, line 17: The `client` can also be used in situation when there is non account connected.
In such situation, it makes sense to create a `Client` beforehand.

It would be nice to be able to pass that instance instead of re-creating a new one from the `ClientConfig`.

Can it be problematic to have multiple instances of `Client`? If not, then probably not a big deal to re-create one when configuring `WebUser`

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+        return WebUser.account
+    }
+
+    public static async getDid() {
+        await WebUser.requireConnection()
+
+        return WebUser.did
+    }
+
+    public static async getPublicProfile(force: boolean = false) {
```
aurelticot, line 71: Worth adding jsdoc mentionning the arg `force` is to refetch the profile instead of potentially using the cache

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        const profile = WebUser.profileConnection
+
+        // load avatar
+        const avatar = await profile.get('avatar')
+
+        // build a cached profile
+        WebUser.profile = {
+            avatarUri: avatar ? avatar.uri : undefined,
+            name: await profile.get('name')
```
aurelticot, line 105: `description` and `country` are missing

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    text: string
+    link?: WebUserMessageLink
+}
+
+export interface WebUserMessageLink {
+    url: string
+    text: string
+}
+
+/**
```
aurelticot, line 33: Worth adding jsdoc mentioning the usage: start by `WebUser.configure(...)`, then `WebUser.connect()`, ...

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    private static connecting: Promise<boolean>
+    private static profileConnection: Profile
+
+    public static configure(config: WebUserConfig) {
+        WebUser.config = config
+    }
+
+    public static async getContext() {
+        await WebUser.requireConnection()
+        return WebUser.context
```
aurelticot, line 56: If we are absolutely sure that `context` is defined when we are connected, then we should narrow the type of context to `Context` instead of `Context | undefined`.

This would greatly help the DX not having to check if `context` is undefined all the time even though the account is connected and the context is there.

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    public static async getAccount() {
+        await WebUser.requireConnection()
+
+        return WebUser.account
```
aurelticot, line 62: Same comment as for the context regarding narrowing the type

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    public static async getDid() {
+        await WebUser.requireConnection()
+
+        return WebUser.did
```
aurelticot, line 68: Same comment as for the context regarding narrowing the type

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+        const isConnected = await WebUser.isConnected()
+        if (!isConnected) {
+            throw new Error('Not connected!')
+        }
+    }
+
+
+    // @todo
+    // public static async requestData() {}
+    // public static async sendData() {}
```
aurelticot, line 249: A common use case for a connected account is to open a `datastore` for the current context.

It will be as simple as:

```ts
public static async openDatastore(schemaURL: string): Promise<Datastore> {
    const context = await WebUser.getContext();
    if (!context) {
        // Unlikely to happen if connected. See comment on narrowing the type
        throw new Error("Context undefined")
    }
    return await context.openDatastore(schemaURL);
}
```

packages/account-web-vault/src/index.ts
```diff
@@ -1,7 +1,9 @@
 import VaultAccount, {hasSession} from "./vault-account"
 import { VaultAccountConfig } from "./interfaces"
+import { WebUser } from "./web-user"
```
aurelticot, line 2: The interfaces are not exported. This can be annoying from a DX point of view.

I would personally simply do:

```ts
export * from "./web-user"
```

Same for the other files.

To me all Types and Interfaces should be externally accessible.

If there are private methods, they should be either kept private in the scope of their file, or in a dedicated file not re-exported by this root `index.ts` (or any other down the hierarchy)

aurelticot (changes requested):

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    text: string
+}
+
+/**
+ * 
+ * @event profileChanged
+ * @event connected
+ * @event disconnected
+ */
+export class WebUser {
```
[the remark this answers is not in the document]
aurelticot: Just adding, it may have some impacts on how to handle such static class with a global/local state like React State or React Context.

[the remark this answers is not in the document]
tahpot: Yes, exactly the rationale.

I’m flexible with either. 

You’re suggesting a singleton because it will play nicer with React?

I originally wrote this within a React app.

packages/account-web-vault/src/web-user.ts
tahpot: Good point.

What if “getClient()” always returned a client object?

packages/account-web-vault/src/index.ts
tahpot: Agreed

packages/account-web-vault/src/web-user.ts
tahpot: Added `getClient()`

packages/account-web-vault/src/web-user.ts
```diff
@@ -0,0 +1,250 @@
...
+    avatarUri: string
+    country?: string
+    description?: string
+}
+
+export interface WebUserConfig {
+    accountConfig: VaultAccountConfig
+    clientConfig: ClientConfig
+    contextConfig: ContextConfig
+    networkConfig: NetworkConnectionConfig
```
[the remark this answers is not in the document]
tahpot: Fixed

packages/account-web-vault/src/web-user.ts
tahpot: Good point

packages/account-web-vault/src/web-user.ts
tahpot: Good point.

packages/account-web-vault/src/web-user.ts
tahpot: Done. Also added `openDatabase`.

packages/account-web-vault/src/web-user.ts
tahpot: Fixed. Renamed to `ignoreCache` to be clearer.

packages/account-web-vault/src/web-user.ts
aurelticot: It means you have to configure WebUser, but I guess it's ok.

If you use WebUser to get the client, it means you need WebUser for something else, so it make sense to configure it.
If you don't want to configure it, it's because you don't need WebUser, so you should just use `new Client(...)`.

packages/account-web-vault/src/web-user.ts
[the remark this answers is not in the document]
tahpot: As discussed, will stick with static as is.

PR #283, opened 2023-02-07
Feature/280 implement sbt client

ITStar10:


packages/vda-sbt-client/src/blockchain/blockchainApi.ts
```diff
@@ -0,0 +1,215 @@
...
+export interface SBTClientConfig {
+    identifier: string;
+    signKey: string;
+    chainNameOrId?: string | number;
+
+    callType: CallType;
+    web3Options: VeridaSelfTransactionConfig | VeridaMetaTransactionConfig;
+}
+
+export interface SBTInfo {
```
tahpot, line 29: @ITStar10 Add documentation that explains the purpose of each attribute

packages/vda-sbt-client/test/blockchain-api.test.ts
```diff
@@ -0,0 +1,198 @@
...
+const createBlockchainAPI = (did: any) => {
+    return new BlockchainApi({
+        identifier: did.address,
+        signKey: did.privateKey,
+        chainNameOrId: "testnet",
+        ...configuration
+    })
+}
+
+describe('vda-sbt-client blockchain api', () => {
```
tahpot, line 22: I'm getting timeouts for all of these. Can you set timeout to be 20 seconds for each test?

I tried setting `this.timeout()`, however it won't work ... which is weird as it works on all our other test suites.

tahpot (changes requested):
See comments

packages/vda-sbt-client/test/blockchain-api.test.ts
tahpot: Is working locally now.

packages/vda-sbt-client/src/blockchain/blockchainApi.ts
ITStar10: Updated.

tahpot (approved):

PR #293, opened 2023-03-05
Fix/292 improve name client

### LINKED ISSUE #292: Feedback on `vda-name-client`
Feedback on [`vda-name-client`](https://github.com/verida/verida-js/blob/main/packages/vda-name-client)

# Clarification

_Note: The first two points are less concerning, knowing this `VeridaNameClient` is embedded in `Client` and would not really be used directly._

- [x] **Clarify the usage of the client**
The [README](https://github.com/verida/verida-js/blob/main/packages/vda-name-client/README.md) make it looks like the `identifier`, `signKey` and `web3Options.privateKey` properties are mandatory when creating an instance while they are not necessary to use `getUsernames()` and `getDid()`. Maybe put a comment  on each line saying they are only required for "writing" on the blockchain, so for `register()` or `unregister()`

- [x] **Clarify the `callType` property in the JS Doc and the README**
It would be helpful to explain a bit more this property, especially as it's a mandatory property for creating an instance of the Client.

- [x] **Clarify the type syntax of the DID returned by `getDid()`**
https://github.com/verida/verida-js/blob/03ed325cca14efca7ca5d42c5128f7fe6f21b0ef/packages/vda-name-client/src/blockchain/blockchainApi.ts#L226
In the `getDid()` method, the returned DID is built with the value of `this.network` after "did:vda:". The type of `this.network` is weak, as based on the `chainNameOrId` property, and following the code, we could have DID like: "did:vda:testnet:0x..." or "did:vda:0x89:0x..." or "did:vda:mainnet:0x..." or "did:vda:0x13881:0x...", but I'm not sure that's what we support? I also had the impression that the mainnet DID would be "did:vda:0x..." with no explicit mention of the network unlike the testnet ones, no?

# DX

- [ ] **Use Conditional Type for the `web3Options` property, if possible**
This property can be of two types `Web3SelfTransactionConfig` or `Web3MetaTransactionConfig`, the JS doc says it depends on the `callType` value. It would be helpful to make a stronger type, probably with a Conditional Type: if `callType` is `web3` then `web3Options` must be `Web3Self`, if it's `gasless` then it's `Web3Meta...`.

- [x] **Define a stronger type for the `chainNameOrId` property in the Client config**
The type is `string | number` and the JS Doc says "chain name or chain id". In the class `VeridaNameClient` it's only assigned to `this.network` on [line 75](https://github.com/verida/verida-js/blob/03ed325cca14efca7ca5d42c5128f7fe6f21b0ef/packages/vda-name-client/src/blockchain/blockchainApi.ts#L75) with a `.toString` and on [line 60](https://github.com/verida/verida-js/blob/03ed325cca14efca7ca5d42c5128f7fe6f21b0ef/packages/vda-name-client/src/blockchain/blockchainApi.ts#L60) without a `.toString`. Then `this.network` is used to get the contract address as a key from a `Record<string, string | null>`, this `CONTRACT_ADDRESS` object could benefit from a stronger key type, ie: the same as `chainNameOrId`.  Seeing the keys of the CONTRACT_ADDRESSES , the values of `chainNameOrId` should be "mainnet", "0x89", "testnet" and "0x13881". It appears the type of `chainNameOrId`, and by extension of `this.network` should be `EnvironmentType`, no?

- [x] **Surround blockchain calls with try/catch to return the specific Error "Failed to get..."**
When using `getDid(“someUsername”)` with a username that doesn’t exist, we get an (expected) Exception but of type: “Call revert Exception: ‘Unregistered’”, which I reckon is coming from deep down when calling the blockchain. In both `VeridaNameClient.getDid()` and `VeridaNameClient.getUsernames()` there is no try/catch surrounding the blockchain methods and they seem to raise Errors which bubble up to the application (the “call revert”, but also the “network doesn’t support ENS”). We should catch them and throw a known/expected Error “Failed to get DID/Username”.

- [x] **Expose more methods from the package [`@verida/client-ts`](https://github.com/verida/verida-js/blob/main/packages/client-ts/src/client.ts)**
Embedding the `VeridaNameClient` in the main Verida `Client` and using it on the different methods is a good idea. But we should expose at least the methods `getDid()` and `getUsernames()` from there. `parseDid()` has a different behaviour than these two methods and its name is not explicit enough.
Note: With the previous point about handling errors wihtin `getDid()` and `getUsernames()`, there is no need for the try/catch within `parseDid()`

# Optimisation

- [x] **Remove unused method argument `network` in the method `interpretIdentifier()`**
This argument is not used and is even reseted to `undefined` at the beginning of the method.
https://github.com/verida/verida-js/blob/03ed325cca14efca7ca5d42c5128f7fe6f21b0ef/packages/vda-name-client/src/blockchain/helpers.ts#L7

- [x] **Use the cache both ways**
There’s a cache but it’s only used to get a DID by a Username, not the other way around. It’s possible to get the list of usernames for a given DID using this cache, like so:
```ts
const searchedDid = "did:vda:...";
const usernames = Object.entries(usernameCache)
  .filter(([, did]) => did === searchedDid)
  .map(([username, did]) => username);
```
For the cache, I would advise using a `Map` rather than an object, it’s cleaner to get/set/delete an entry.

# Bugs

- [x] **All calls to `getUsernames()` return an Error**
When calling `nameClient.getUsernames(did)`, I always get an Error “network doesn’t support ENS”, for every cases I tested whether the DID doesn’t exist, exists but doesn’t have usernames, or exists with usernames.

tahpot:
Closes #292

### DIFF
--- a/packages/vda-name-client/src/blockchain/blockchainApi.ts
+++ b/packages/vda-name-client/src/blockchain/blockchainApi.ts
@@ -194,7 +194,7 @@ export class VeridaNameClient {
                         return []
                     }
 
-                    throw new Error(`Failed to get usernames for DID (${did}): ${response.reason}`)
+                    throw new Error(response.reason)
                 }
 
                 return response.data
@@ -214,6 +214,7 @@ export class VeridaNameClient {
 
     /**
      * Return the DID address for a given username
+     * 
      * @param username username registered by {@link register}
      * @returns DID address
      */
@@ -228,14 +229,14 @@ export class VeridaNameClient {
                 response = await this.vdaWeb3Client.findDID(username)
 
                 if (response.success !== true) {
-                    throw new Error(`Failed to locate the DID for username: ${username}`)
+                    throw new Error(`Not found`)
                 }
 
                 response = response.data
             } else {
                 response = await this.contract!.callStatic.findDID(username)
                 if (!response) {
-                    throw new Error(`Failed to locate the DID for username: ${username}`)
+                    throw new Error(`Not found`)
                 }
             }
 
--- a/packages/vda-name-client/src/blockchain/blockchainApi.ts
+++ b/packages/vda-name-client/src/blockchain/blockchainApi.ts
@@ -165,7 +165,8 @@ export class VeridaNameClient {
 
     /**
      * Get the username list of a DID address
-     * @param did DID address to get usernames
+     * 
+     * @param did DID to lookup the username for
      * @returns username list 
      */
     public async getUsernames(did: string): Promise<string[]> {
@@ -174,6 +175,15 @@ export class VeridaNameClient {
         if (didAddress.match('did')) {
             const { address } = explodeDID(did)
             didAddress = address
+        }
+
+        // Lookup usernames from cache
+        const usernames = Object.entries(this.usernameCache)
+            .filter(([, existingDid]) => existingDid === did)
+            .map(([username, existingDid]) => username)
+        
+        if (usernames.length) {
+            return usernames
         }
 
         try {
--- a/packages/vda-name-client/src/blockchain/blockchainApi.ts
+++ b/packages/vda-name-client/src/blockchain/blockchainApi.ts
@@ -5,12 +5,14 @@ import {
 import {
     Web3SelfTransactionConfig,
     Web3MetaTransactionConfig,
-    Web3CallType
+    Web3CallType,
+    EnvironmentType
 } from '@verida/types'
 import { ethers, ContractFactory } from "ethers";
-import { getContractInfoForNetwork } from "./config";
-import { getVeridaSignWithNonce, interpretIdentifier } from "./helpers";
+import { getContractInfoForNetwork, RPC_URLS } from "./config";
+import { getVeridaSignWithNonce } from "./helpers";
 import { JsonRpcProvider } from '@ethersproject/providers';
+import { explodeDID } from '@verida/helpers'
 
 /**
  * Interface for vda-name-client instance creation. Same as VDA-DID configuration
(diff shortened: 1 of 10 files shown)


packages/vda-name-client/src/blockchain/blockchainApi.ts
```diff
@@ -21,12 +23,11 @@ import { JsonRpcProvider } from '@ethersproject/providers';
...
-    signKey?: string;
-    chainNameOrId?: string | number;
-
-    callType: Web3CallType;
-    web3Options: Web3SelfTransactionConfig | Web3MetaTransactionConfig;
+    network: EnvironmentType
+    did?: string
+    signKey?: string
+    callType?: Web3CallType
+    web3Options?: Web3SelfTransactionConfig | Web3MetaTransactionConfig
```
aurelticot, line 30: For the conditional type, it'd actually be a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

```ts
export type NameClientConfig = {
  network: EnvironmentType
  did?: string
  signKey?: string
} & (
  | {
      callType?: "web3";
      web3Options?: Web3SelfTransactionConfig;
    }
  | {
      callType?: "gasless";
      web3Options?: Web3MetaTransactionConfig;
    }
);
```

aurelticot (approved):
Looks good and approved.

I just added an example for the type of `web3Options`.

And a couple of comments for the errors thrown in `getUsernames` and `getDID`.

packages/vda-name-client/src/blockchain/blockchainApi.ts
```diff
@@ -143,57 +146,69 @@ export class VeridaNameClient {
...
+
+        try {
+            if (this.vdaWeb3Client) {
+                response = await this.vdaWeb3Client.getUserNameList(didAddress)
+                if (response.success !== true) {
+                    if (response.reason == 'No registered DID') {
+                        return []
+                    }
+
+                    throw new Error(`Failed to get usernames for DID (${did}): ${response.reason}`)
```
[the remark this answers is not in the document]
tahpot: Good point

packages/vda-name-client/src/blockchain/blockchainApi.ts
tahpot: I get a syntax error with that, so have left for now.

Happy to accept a PR if you can get it to work.

PR #295, opened 2023-03-06
Fix: SBT signature verification issue 294

### LINKED ISSUE #294: `vda-sbt-client` signature verification fails
When attempting to sign a SBT request, the following error is returned:

`Failed to claim SBT: Data is not signed by a valid signing DID`

tahpot:
Closes #294.

Add a Polygon private key to `test/utils.ts` L24.


tahpot:
Merging, although two unit tests are broken which @ITStar10 will investigate.

PR #318, opened 2023-04-27
[vda-common] Update contract name

ITStar10:


ITStar10:
## Updated
Following packages are updated. Their package versions are also upgraded:
- vda-common
- vda-did-resolver
- vda-did
- vda-sbt-client

## Test
### VDA-XX packages
Following packages are passed in tests:
- vda-did
- vda-did-resolver
- vda-name0client
- vda-sbt-client

### Other packages
image

PR #335, opened 2023-07-06
Fix/332 `WebUser` using previous identity after disconnecting and connecting with a different one

aurelticot:
The `disconnect` method was not cleaning the `profileConnection` instance property that is used to get the profile.

### DIFF
--- a/packages/web-helpers/CHANGELOG.md
+++ b/packages/web-helpers/CHANGELOG.md
@@ -1,3 +1,8 @@
+2023-xx-xx (c2.3.x)
+-------------------
+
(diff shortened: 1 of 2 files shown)


tahpot (approved):
Thanks for this.

Can you also add a placeholder CHANGELOG entry with no date or release number

aurelticot:
> Can you also add a placeholder CHANGELOG entry with no date or release number

Done.
Not sure how the release works without the `develop` branch. Is it ok to merge on `main`?

tahpot (approved):
Looks good.

Issue #336, opened 2023-07-06
Refactor `WebUser` - Breaking Changes


aurelticot:
Most the following is introducing BREAKING CHANGES!

```[tasklist]
### Tasks
- [x] Move client instance creation in constructor so property is never undefined
- [x] Create dedicated auto connect method if has sessions (to avoid `isConnect` to be a Promise)
- [x] Check `context`, `account` and `did` instance property in `isConnect`
- [x] Remove returning Promises when not needed
- [x] Fix `WebUserProfile.avatarUri` not undefined in interface definition while it can be
- [x] Optimise imports with `import type`
- [x] Simplify TypeScript syntax
```

The main (breaking) change is about functions returning Promises while it's not needed, which is a bit annoying at usage.

`getDid`, `getContext`, `getAccount` were returning Promises because of `requireConnection` being async, because of `isConnected` being async, because of the auto-connnect.
By removing the auto-connect from `isConnected`, moving it into a dedicated function, we remove all the subsequent Promises.

`getClient` returned a Promise for no reasons.

PR #337, opened 2023-07-06
Feature/336 - Refactor `WebUser` - Breaking Changes

aurelticot:
This is introducing **Breaking Changes!**

Check #336 for details


packages/web-helpers/src/WebUser.ts
```diff
@@ -1,11 +1,11 @@
...
+import { Client, Network, type Context } from '@verida/client-ts';
 import { VaultAccount, hasSession } from '@verida/account-web-vault';
 import { EventEmitter } from 'events'
-import { IProfile, ClientConfig, ContextConfig, MessageSendConfig, IDatastore, IDatabase, DatabaseOpenConfig, DatastoreOpenConfig, AccountVaultConfig } from '@verida/types';
+import type { IProfile, ClientConfig, ContextConfig, MessageSendConfig, DatabaseOpenConfig, DatastoreOpenConfig, AccountVaultConfig } from '@verida/types';
 
 export interface WebUserProfile {
     name?: string
-    avatarUri: string
+    avatarUri?: string
```
tahpot, line 8: This is not a required field in the profile schema

packages/web-helpers/src/WebUser.ts
```diff
@@ -234,69 +289,39 @@ export class WebUser extends EventEmitter {
...
+     * Open a datastore owned by this user.
+     *
+     * @param {string} schemaURL
+     * @param {DatastoreOpenConfig} config
+     * @returns A Promise that will resolve to the datastore instance.
      */
-    public async openDatastore(schemaURL: string, config?: DatastoreOpenConfig): Promise<IDatastore> {
-        const context = await this.getContext()
-        return await context.openDatastore(schemaURL, config)
+    public openDatastore(schemaURL: string, config?: DatastoreOpenConfig) {
```
tahpot, line 306: Must be async

packages/web-helpers/src/WebUser.ts
```diff
@@ -234,69 +289,39 @@ export class WebUser extends EventEmitter {
...
+     * Open a database owned by this user.
+     *
+     * @param {string} databaseName
+     * @param {DatabaseOpenConfig} config
+     * @returns A Promise that will resolve to the database instance.
      */
-    public async openDatabase(databaseName: string, config?: DatabaseOpenConfig): Promise<IDatabase> {
-        const context = await this.getContext()
-        return await context.openDatabase(databaseName, config)
+    public openDatabase(databaseName: string, config?: DatabaseOpenConfig) {
```
tahpot, line 318: Must be async

packages/web-helpers/src/WebUser.ts
```diff
@@ -185,43 +238,45 @@ export class WebUser extends EventEmitter {
...
-     * @param {*} linkText 
+     * Send a generic message to a user's inbox (accessible from the Verida Wallet).
+     *
+     * @param {string} did
+     * @param {WebUserMessage} message the message definition
      */
-    public async sendMessage(did: string, message: WebUserMessage): Promise<void> {
-        const context = await this.getContext()
-        const messaging = await context!.getMessaging()
+    public async sendMessage(did: string, message: WebUserMessage) {
```
tahpot, line 277: Must be async

packages/web-helpers/src/WebUser.ts
```diff
@@ -171,12 +222,14 @@ export class WebUser extends EventEmitter {
...
+            const profile = await this.getPublicProfile()
+
             this.account = account
-            this.context = context!
+            this.context = context
             this.did = did
 
-            const profile = await this.getPublicProfile()
-            this.client = context!.getClient()
+            // Question: Do we have to get overwrite the client created initially in the constructor by the client from context? (knowing we created the context from this client while connecting)
```
tahpot, line 231: Yes, creating the context uses the account object which is then used to create a client that contains that account object.

tahpot (approved):
Reveiwed. Looks good. See comments.

Please test very thoroughly.

packages/web-helpers/src/WebUser.ts
aurelticot: Yes, that's why I put it optional here, it was not.

packages/web-helpers/src/WebUser.ts
aurelticot: It is async, we just don't need the syntax as it's inferred by the return of the Promise from `messaging.send`.

packages/web-helpers/src/WebUser.ts
aurelticot: It is async, we just don't need the syntax as it's inferred by the return of the Promise from `context.openDatastore`.

packages/web-helpers/src/WebUser.ts
aurelticot: It is async, we just don't need the syntax as it's inferred by the return of the Promise from `context.openDatabase`.

packages/web-helpers/src/WebUser.ts
aurelticot: Oh, ok, because `Network.connect` which takes the config of the Client, not a Client instance

packages/web-helpers/src/WebUser.ts
tahpot: I'm confused by this, see Discord thread. We can summarise that conversation here once concluded.

PR #339, opened 2023-07-14
Closes 338. Map the DID string to the correct DID format

nick-verida:


packages/vda-did-resolver/src/getDIDs.ts
```diff
@@ -28,13 +28,21 @@ export async function getDIDs(network: string, startIndex: number = 0, count: nu
...
     }
     
-    let data;
+    let data: string[] | undefined = undefined;
     try {
-        data = await contract.callStatic.getDIDs(startIndex, count)
+        data = (await contract.callStatic.getDIDs(startIndex, count)) as string[];
+        data.map((did) => {
+            // map the raw string into the correct canonoical DID format
+            return `did:vda:${network}:${did}`;
```
aurelticot, line 44: Do we have a centralised util function for this formatting?

packages/vda-did-resolver/src/getDIDs.ts
```diff
@@ -10,7 +10,7 @@ import { activeDIDCount } from "./activeDIDCount";
  * @param startIndex Index to start
  * @param count Total number of results to fetch
  */
-export async function getDIDs(network: string, startIndex: number = 0, count: number=20, mostRecent: boolean=true): Promise<string[]> {
+export async function getDIDs(network: string, startIndex: number=0, count: number=20, mostRecent: boolean=true): Promise<string[]> {
```
aurelticot, line 13: Consider having a stronger type on `network`

packages/vda-did-resolver/src/getDIDs.ts
```diff
@@ -28,13 +28,21 @@ export async function getDIDs(network: string, startIndex: number = 0, count: nu
...
     
-    let data;
+    let data: string[] | undefined = undefined;
     try {
-        data = await contract.callStatic.getDIDs(startIndex, count)
+        data = (await contract.callStatic.getDIDs(startIndex, count)) as string[];
+        data.map((did) => {
+            // map the raw string into the correct canonoical DID format
+            return `did:vda:${network}:${did}`;
+        })
```
aurelticot, line 45: `map` doesn't mutate the array, so the `return data` is not modified as intended.

I suggest the following:

```ts
    // Move the try at the very top as the catch re-throws a generic error with a cause
    try {

        // ... 

        const data = (await contract.callStatic.getDIDs(startIndex, count)) as string[];
        // TODO: Consider validating `data` before processing and then removing the `as string[]` assertion
        // In the meantime, have a fall back with `data || []`, not the best though
        return (data || []).map((did) => `did:vda:${network}:${did}`)
    } catch (error: unknown) {
        throw new Error('Failed to get list of active DIDs', { cause: error });
    }
}
```

packages/vda-did-resolver/src/getDIDs.ts
```diff
@@ -10,7 +10,7 @@ import { activeDIDCount } from "./activeDIDCount";
  * @param startIndex Index to start
  * @param count Total number of results to fetch
  */
-export async function getDIDs(network: string, startIndex: number = 0, count: number=20, mostRecent: boolean=true): Promise<string[]> {
+export async function getDIDs(network: string, startIndex: number=0, count: number=20, mostRecent: boolean=true): Promise<string[]> {
     const rpcUrl = RPC_URLS[network]
```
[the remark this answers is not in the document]
nick-verida: agree

packages/vda-did-resolver/src/getDIDs.ts
nick-verida: Not that I've seen anywhere

packages/vda-did-resolver/src/getDIDs.ts
nick-verida: Agree. Fixing

packages/vda-did-resolver/src/getDIDs.ts
nick-verida: Agree

aurelticot (approved):
LGTM

PR #345, opened 2023-09-04
Don't force global JSON.stringify to be sortable

### LINKED ISSUE #343: Remove JSON.stringify global scope hot replacement with `json.sortify`
Hello everybody,

The aforementioned [issue](https://github.com/verida/verida-js/blob/c94b95de687c64cc776652602665bb45a327dfb6/packages/encryption-utils/src/index.ts#L10).

For context:

This technique renders the deterministic nature of JSON.stringify unusable, since keys are not preserved, rather than sorted alphabetically.

Having that in mind, the issue is that the hot replacement is done in global scope, i.e. importing a module that utilises the encryption-utils declarations _or_ bundling altogether is sufficient to accidentally end up with non-deterministic behaviour.

An example is:

- Deserialising a string to a struct / object.
- Process / alter state.
- Serialise + sort object keys by default. <-- Point of non-deterministic outcome, if integrity hashes are in place and / or JSON hashing <> buffer conversion is in place.

Suggestion:

I'd suggest swapping out + enhancing the global JSON object with a `sortify` method and / or serialising + sorting in 2 distinct steps.
Otherwise, a simple utility function could be sufficient, depending on you internal usage of key sorting on serialisation.

Let us know your thoughts.

tahpot:
Closes #343

### DIFF
--- a/packages/encryption-utils/src/index.ts
+++ b/packages/encryption-utils/src/index.ts
@@ -113,7 +113,7 @@ export default class EncryptionUtils {
         if (data instanceof Uint8Array) { // this also covers `Buffer`
             data = utils.hexlify(data) // the hashing method that follows also tries to serialize strings, except when they are hex strings
         } else if (typeof data === 'object') {
-            data = JSON.stringify(data)
+            data = JSONSortify(data)
         }
 
         const messageHashBytes = EncryptionUtils.hashBytes(data)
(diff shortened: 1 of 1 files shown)


aurelticot (changes requested):
1 unit test for the `encryption-utils` failed.

There's a remaining `JSON.stringify` in the file.

https://github.com/verida/verida-js/pull/345/files#diff-6d7a75184dbadd43e1b42eeddc4092399280c246931276f0c30a13d6e1bed36cR116

tahpot:
Oops. Fixed.

cc @aurelticot

aurelticot (approved):

PR #354, opened 2023-10-29
Feature/353 support replicating contexts

### LINKED ISSUE #353: Support replication of application context
Need to support replicating all the databases for an application context from one storage node to another storage node.

This is useful for migrating data from `testnet` to `mainnet`. As this is a once off operation, so create it in the helpers library, `@verida/helpers`.

The overarching flow will be:

1. Open `source` storage context and `destination` storage context
2. Loop through all databases in `source` storage context
3. For each database:
 a. Open the same database in the `destination` storage context
 b. Replicate all data
 c. Raise an event that x/y databases have been syncronized

tahpot:
Closes #353

### DIFF
--- a/packages/client-ts/src/context/context.ts
+++ b/packages/client-ts/src/context/context.ts
@@ -426,6 +426,20 @@ class Context extends EventEmitter implements IContext {
     return database
   }
 
+  public async deleteDatabase(databaseName: string) {
+    if (!this.account) {
+      throw new Error(`Unable to delete database. No authenticated user.`);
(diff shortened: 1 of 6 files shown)


tahpot:
See tests on how the events are emitted to notify the progress of the migration, which will be helpful to display in the user interface:

https://github.com/verida/verida-js/pull/354/files#diff-d94d706630a3f377d01588c3498bb5c8e9e091ff75a89fbf77735fa275e4cb9eR76

PR #379, opened 2023-12-19
Add the missing `environment` property to `AccountVaultRequest`

aurelticot:
@tahpot I added the missing property.

I put it as optional for backward compatibility.

But it should be required. In that case, it's be in a Major version 3.0.0, which would make sense with the Mainnet being the default, which is a breaking change.


aurelticot:
Misunderstanding on where the `environment` is to be set. It's not in `AccountVaultRequest` but in `AccountVaultConfig` where it is already available

PR #382, opened 2023-12-21
Fix migration opening database with wrong config

aurelticot:
Both the source database and destination database were passed the same config object
```ts
const config = {
    permissions: sourceDbInfo.permissions,
    verifyEncryptionKey: false
}
```
Problem was the `...Context.openDatabase(databaseName, config)` mutates the config object to add the DID if missing or use it if available.
So the `sourceContext.openDatabase` would add the source DID to the shared config object and when opening the destination database, the same DID from the source was used instead of the destination DID.


aurelticot:
Also added a 'migrated' event even if the database is skipped otherwise it messes up the progress.

tahpot (approved):

PR #408, opened 2024-03-22
Feature/407 update vda node xxx packages

ITStar10:
- Updated the `vda-node-manager` and `vda-node-client` functions
- Update the `StorageNodeRegistry` contract address and abi in the `vda-common`

### DIFF
--- a/packages/vda-node-manager/test/user_read.test.ts
+++ b/packages/vda-node-manager/test/user_read.test.ts
@@ -308,6 +308,12 @@ describe('vda-node-manager read only tests', () => {
                     assert.ok(err.message.match('Failed to get nodes by country'), 'Invalid page number');
                 }
 
+                try {
+                    await blockchainApi.getNodesByCountryCode(countryCodes[0], 101, 1);
+                } catch(err) {
+                    assert.ok(err.message.match('Failed to get nodes by country'), 'Invalid page number');
+                }
+
                 // Success
                 const result = await blockchainApi.getNodesByCountryCode(countryCodes[0], 1, 2);
                 assert.ok(result.length === 1, "Get nodes by country code");
--- a/packages/vda-node-manager/test/user_read.test.ts
+++ b/packages/vda-node-manager/test/user_read.test.ts
@@ -288,10 +288,29 @@ describe('vda-node-manager read only tests', () => {
     
                 for (let i = 0; i < countryCodes.length; i++) {
                     const countryNodes = nodes.filter(item => item.countryCode === countryCodes[i]);
-                    const result = await blockchainApi.getNodesByCountryCode(countryCodes[i]);
+                    const result = await blockchainApi.getNodesByCountryCode(countryCodes[i], 100, 1);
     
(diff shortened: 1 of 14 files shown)


packages/vda-node-manager/test/user_write.test.ts
```diff
@@ -39,6 +39,7 @@ describe('vda-node-manager read and write tests', () => {
         registeredCentreIds = await addInitialData(configuration);
     })
 
+    /*
```
tahpot, line 42: Remove

packages/vda-node-manager/test/user_read.test.ts
```diff
@@ -330,29 +372,33 @@ describe('vda-node-manager read only tests', () => {
...
-                const result = await blockchainApi.getNodesByRegionCode(removedregionCode, EnumStatus.removing);
+                let result = await blockchainApi.getNodesByRegionCode(removedregionCode, 100, 1, EnumStatus.removing);
                 assert.ok(result.length > 0, "Get 'removing' nodes by region code and status");
+
+                // Check pagination
+                result = await blockchainApi.getNodesByRegionCode(regionCodes[0], 1, 2, EnumStatus.active);
+                assert.ok(result.length === 1, "Get 'active' nodes by region code");
             })
 
             it("Get nodes by status", async () => {
```
tahpot, line 391: For all these pagination tests, can you also add a test to show the behaviour if you request results outside of the number of nodes in the contract.

ie: there are 100 nodes, but you request nodes from 150-180.

Also, what happens if you request 10,000 nodes? Is that too many? Do we need to limit the page size?

tahpot (changes requested):
See comments

packages/vda-node-manager/test/user_write.test.ts
ITStar10: It's used in the `Data Centre Test`.
Pushed a new commit

packages/vda-node-manager/test/user_read.test.ts
ITStar10: 1. When there are 100 nodes and you requested nodes from 150-180, it will returns an empty array.
2. We need to limit the page size

tahpot: Have you limited the page size?

tahpot (changes requested):
With the CHANGELOG.md files, please specify no date or version number. I will add these when the changes are actually published to npm, along with other changes.

packages/vda-node-manager/test/user_read.test.ts
ITStar10: Pls check following functions:
```
getNodesByCountryCode()
getNodesByRegionCode()
getNodesByStatus()
```

ITStar10: Pls check following files:
`vda-node-client/src/blockchain/blcochainApi.ts`
`vda-node-manager/src/blockchain/ownerApi.ts`

tahpot (approved):

PR #410, opened 2024-03-27
Implement getPublicProfile with tests

tahpot:


packages/client-ts/src/client.ts
```diff
@@ -208,6 +210,39 @@ class Client implements IClient {
...
   }
 
+  public async getPublicProfile(did: string,
+    contextName: string,
+    profileName: string = "basicProfile",
+    fallbackContext: string | null = "Verida: Vault",
+    useCache: boolean = true): Promise<ProfileDocument> {
+    if (this.config.readOnlyDataApiUri) {
+      // Try to fetch the profile from our data API
+      const fetchUri = `${this.config.readOnlyDataApiUri}/${did}/${contextName}/profile_public/${profileName}?ignoreCache=true`
```
andy-verida, line 220: `ignoreCache=true` is this params specified by the API consumer

packages/client-ts/src/client.ts
```diff
@@ -208,6 +210,39 @@ class Client implements IClient {
     return this.config
   }
 
+  public async getPublicProfile(did: string,
+    contextName: string,
+    profileName: string = "basicProfile",
+    fallbackContext: string | null = "Verida: Vault",
+    useCache: boolean = true): Promise<ProfileDocument> {
```
aurelticot, line 217: As it's an optional parameter and the default is to use cache, the parameter should actually be `ignoreCache = false`.

Because, from a DX perspective, not defining a parameter usually means it's falsy: `ignoreCache = undefined` is equivalent to `ignoreCache = false` and `useCache = undefined` is equivalent to `useCache = false`. Because we want to use the cache by default, we can just flip the name and meaning of the parameter `userCache` to `ignoreCache`.

Also in that case, the data URL specify a `ignoreCache` param, so it's more straightforward to pass `ignoreCache=${ignoreCache}` rather than `ignoreCache=${!useCache}`.

packages/client-ts/src/client.ts
```diff
@@ -208,6 +210,39 @@ class Client implements IClient {
...
+    if (this.config.readOnlyDataApiUri) {
+      // Try to fetch the profile from our data API
+      const fetchUri = `${this.config.readOnlyDataApiUri}/${did}/${contextName}/profile_public/${profileName}?ignoreCache=true`
+
+      try {
+        const response = await axios.get(fetchUri)
+        return <ProfileDocument> response.data
```
aurelticot, line 224: Nothing to do, but a remark we will have to make some proper validations and typings throughout the libraries rather assertions.

aurelticot (changes requested):

packages/client-ts/src/client.ts
tahpot: Should be resolved.

packages/client-ts/src/client.ts
tahpot: Agreed.

packages/client-ts/src/client.ts
```diff
@@ -208,6 +210,39 @@ class Client implements IClient {
     return this.config
   }
 
+  public async getPublicProfile(did: string,
+    contextName: string,
+    profileName: string = "basicProfile",
+    fallbackContext: string | null = "Verida: Vault",
```
aurelticot, line 216: Just to be sure.
It is intentional to allow `null` for the fallbackContext?
So that if set to `null`, we don't use the fallback, but if we let it `undefined`, we will fallback to the Vault, right?

aurelticot (approved):
LGTM

packages/client-ts/src/client.ts
andy-verida: I known about this. Yes, this is the intention.

aurelticot: Ok, thanks.
I prefer to make sure because using `null` doesn't apply the default value.

PR #414, opened 2024-04-05
Update to node 18. Closes #413

### LINKED ISSUE #413: Update to node 18 and Lerna 7+
Node 14 is out of maintence in May. We've been using node 18 (`lts/hydrogen`) for most apps for a while so migrate to that. 

This also affects the lerna version

nick-verida:


nick-verida:
Oops, this should be against develop

Issue #402, opened 2024-02-07
Redefine the engine requirement and the typescript config


aurelticot:
> [!CAUTION]
> Changing the engine requirement and the build target is a **breaking change!**

The engine requirement in `package.json` is a result of the compilation of the code, which itself is defined based on the environments the SDK is expected to run.

|  | ES2020 | ES2021 | ES2022 |
| --- | --- | --- | --- |
| Node 14 (unmaintained) | No | Yes | Yes |
| Node 16 (unmaintained) | Yes | Yes | Yes |
| Node 18 (maintenance) | Yes | Yes | Yes |
| Node 20 (current LTS) | Yes | Yes | Yes |
| Browser | 93% of all users | 92% of all users | 85-91% of all users |
| React Native JSC | ? | ? | ? |
| React Native Hermes | ? | ? | ? |

By seeing the current support, ES2020 is widespread enough for browser support and Node under maintenance. By choosing ES2020, we can set the engine requirement to Node 16 to still allow projects stuck with that version.

We'd still have to check the dependencies are compatible (if not, look at updating them)

### Recommended values:

- Typescript `target`: `es2020` (would result in a more optimised code that the current `es5`)
- Typescript `lib`: `es2020` (same as `target`)
- Typescript `module`: `commonjs`
- package.json `engine`: `node >= 16` (breaking change)

### Other recommendations:

- All packages must have the same `target` in their respective `tsconfig`
  - Suggest using a base `tsconfig` extended by the package one, so the common configuration is defined once in the base
- All packages must defined the same `engine` requirements in their respective `package.json`
  - To be fair, it's already the case
-Run the CI with a matrix of Node versions
  - Testing and building from the minimum version of Node supported (engine requirement) to the current LTS, and even to the latest: Node 16, 18, 20, 21.
  - If not, at least the version defined in `.nvmrc`
- `.nvmrc` is only used for dev, test and build, not runtime
  - Recommendation to use the LTS, currently Node 20 (`lts/iron`)
    - But only if the CI is run with a matrix checking the lowest version support
  - If CI not run with a matrix, `.nvmrc` should be aligned with the `engine` requirements
    - Ensure we develop, test and build with the lowest supported version
  - Use the name instead of a defined version: `lts/hydrogen` for node 18
- The actual build being published should be done with the version defined in `.nvmrc`

PR #415, opened 2024-04-05
update to node 18 and update lerna. Closes #413

### LINKED ISSUE #413: Update to node 18 and Lerna 7+

nick-verida:


aurelticot:
It's not as simple as updating `.nvmrc`, please see details in https://github.com/verida/verida-js/issues/402

nick-verida:
While all the thing in #402 are probably correct we've been running things under node 18 for a while (not node 20). Also for a library it seems to make sense to move to the lowest possible version so we don't lockout other projects that are on node 18.

We have to do _something_ and this seems extremely safe.

aurelticot:
> While all the thing in #402 are probably correct we've been running things under node 18 for a while (not node 20). Also for a library it seems to make sense to move to the lowest possible version so we don't lockout other projects that are on node 18.
> 
> We have to do _something_ and this seems extremely safe.

I disagree.

This PR only updates `.nvmrc` which is only used for dev, tests and build. The `engine` requirements is what's important and what constrain other projects.

The `engine` is currently defined with `node >=14` but our tests are run with whatever is defined in `.nvmrc` so if it is node 18, there is no guarantee these libraries will work on node 14 to 17, so, no, it is not extremely safe!

It is reasonable to think no projects should run on node lower than 18 (actually even 20), so if we "have to do _something_", it is about upgrading the `engine` requirement. It's a breaking change and the purpose of #402, which highlight a broader consideration, with changes across `package.json`, `tsconfig.json` and `.nvmrc`. That's my point, it's not as simple as only updating `.nvmrc`.

nick-verida:
> The engine is currently defined with node >=14 but our tests are run with whatever is defined in .nvmrc so if it is node 18, there is no guarantee our libraries will work on node 14 to 17. So, no, it is not extremely safe!

Yes this is a good point. But it's still better than now where the tests run in node 14 (which - again - is about to be out of maintenance)

I agree we should do all the things in #402 (although I think I'd still argue for node 18 instead of node 20). 

I don't have time to do all of them now though, and by doing this change at least we'd test under the same node version we run things on!

nick-verida:
(I was going to update the version in package.json too but clearly we don't want to do that since we still want to allow people to use earlier versions if they choose)

aurelticot:
> I agree we should do all the things in https://github.com/verida/verida-js/issues/402 (although I think I'd still argue for node 18 instead of node 20).

#402 suggests node 20 in `.nvmrc` **only** if we run the tests on multiple node version, otherwise `.nvmrc` should be aligned with the `engine` requirement.

> (I was going to update the version in package.json too but clearly we don't want to do that since we still want to allow people to use earlier versions if they choose)

#402 suggests the `engine` requirement to be `node >= 16`, to take into account that 14 is obsolete but 16 is still maintained and we want to support projects still using it.

> by doing this change at least we'd test under the same node version we run things on!

"... we run things on" but **we** are not the only one running things on. At this stage it's fair to say we are the main consumer of the SDK and we know who use it and can inform them there's a lack of test converage on earlier supposedly supported node version. But I personally don't want this to be a precedent, so by principle I will stay against this PR as it is. I'll leave @tahpot with the decision to merge it.

nick-verida:
This now fixes the docs and updates lerna too

README.md
```diff
@@ -39,7 +39,7 @@ It's not possible to add dependencies between monorepo packages using yarn (ie:
 Unpublished dependencies between monorepo packages can be linked by:
 
 - Manually adding the expected dependency to `package.json` (ie: `@verida/encryption-utils`)
-- Run `npx lerna bootstrap` in the root directory of this project
+- Run `yarn install` in the root directory of this project
```
tahpot, line 42: @nick-verida This change is incorrect, please revert.

README.md
```diff
@@ -49,7 +49,7 @@ Build everything:
 In the root directory:
 
 ```
-npx lerna bootstrap
+yarn install
```
tahpot, line 52: @nick-verida Is it wort adding `nvm use` here?

tahpot (changes requested):

README.md
nick-verida: It is in the section Developer Notes -> Node Version

README.md
nick-verida: @tahpot yes, `lerna bootstrap` has been replaced:

> Replace lerna bootstrap with npm install (or yarn/pnpm). If you are already performing your package manager's install command somewhere in your workflow before where you had previously called lerna bootstrap, then you can just delete it instead. 

https://lerna.js.org/docs/legacy-package-management

aurelticot (changes requested):
The `helpers/dist` folder has been commited by mistake

Issue #411, opened 2024-04-04
Refactor how DIDs and Verida networks work


tahpot:
Key requirements:

1. A DID document needs to become completely independent from the Verida network
2. A single DID document can be linked to contexts across different Verida networks
3. Non-Verida DIDs can be used on the Verida network

We will support a new DID method format for VDA DIDs: `did:vda:[anchored-blockchain]:0x…`. Here are examples:

1. `did:vda:amoy:0x..`
2. `did:cheqd:0x..`
3. `did:base:0x..`
4. `did:polygonid:polygon:main:0x`
5. `did:polygonid:polygon:mumbai:0x`

Anchored blockchains will be a human readable name, that is effectively a lookup within the Verida DID resolver:

1. `polamoy`: Polygon Amoy
2. `polpos`: Polygon PoS
3. `polzkevm`: Polygon zkEVM
4. `base`: Coinbase Mainnet
5. `basesepolia`: Coinbase Sepolia

We need to add `&network=` to DID Document / serviceEndpoint / id which specifies the Verida network associate with that context.

The Verida SDK will support three Verida networks:

1. `devnet`
2. `myrtle`
3. `banksia`
4. `mainnet` (for backwards compatibility, is actually `myrtle`)

tahpot:
I have hit a mixture of problems here when dealing with existing `mainnet` DIDs.

I thought the best approach would be to migrate `mainnet` DIDs to `polpos` DIDs, and update the DID document.

However the `did` is very important in the protocol and was assumed to never change.

Here are key areas the `did` is used:

1. Generating a unique database hash from a database name (`dbHash = [did.toLowerCase(), contextName, databaseName,].join("/")`)
2. Signing a consent message to authenticate with storage nodes

It is now clear that migrating `mainnet` to `polpos` DIDs will not work, so instead will need to support some type of virtual mapping.

PR #423, opened 2024-04-22
WIP to support context information in DID document

tahpot:


tahpot:
Merging so work can continue in #411

PR #433, opened 2024-05-24
update vc functions to add options - especially removeOriginalFields=false

icetrust0212:


packages/verifiable-credentials/src/credentials.ts
```diff
@@ -289,13 +309,16 @@ export default class Credentials {
 		}
 	}
 
-	public async createCredentialJWT({ subjectId, data, schema, context, payload, options }: CreateCredentialJWT): Promise<string> {
+	public async createCredentialJWT(
+		{ subjectId, data, schema, context, payload, options }: CreateCredentialJWT,
```
aurelticot, line 313: No need to set the default valut of the credentialOptions here, it will eventually get the defaults from `this.createVerifiableCredential`

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -92,6 +101,7 @@ export default class Credentials {
 	public async createVerifiablePresentation(
 		vcJwts: string[],
 		context: IContext,
+		options: CreatePresentationOptions = { removeOriginalFields: false },
 		issuer?: any,
```
aurelticot, line 105: `options` should be after `issuer` for backward compatibility

packages/verifiable-credentials/src/credentials.ts
```diff
@@ -127,10 +142,15 @@ export default class Credentials {
 	 * @param {string} didRegistryEndpoint
 	 * @param {string} currentDateTime to allow the client to migrate cases where the datetime is incorrect on the local computer
 	 */
-	public async verifyCredential(vcJwt: string, resolverConfig?: Web3ResolverConfigurationOptions, currentDateTime?: string): Promise<any> {
+	public async verifyCredential(
+		vcJwt: string,
+		options: VerifyCredentialOptions = {removeOriginalFields: false},
+		resolverConfig?: Web3ResolverConfigurationOptions,
+		currentDateTime?: string,
```
aurelticot, line 149: `options` should be last for backward compatibility

packages/verifiable-credentials/src/sharing-credential.ts
```diff
@@ -24,10 +25,11 @@ export default class SharingCredential {
 	/**
 	 * Method to encrypt and issue credential
 	 * @param didJwtVc
+	 * @param options
 	 * @returns
 	 */
 
-	async issueEncryptedPresentation(didJwtVc: any): Promise<any> {
+	async issueEncryptedPresentation(didJwtVc: any, options: CreatePresentationOptions = {removeOriginalFields: false}): Promise<any> {
```
aurelticot, line 32: No need to set the default value for `options` as it will eventually use the defauls from `this.issuePublicPresentation`

aurelticot (changes requested):
A few changes but otherwise looks ok.

Only static review though, I'm not able to test it.

packages/verifiable-credentials/src/credentials.ts
icetrust0212: OK, Done

packages/verifiable-credentials/src/credentials.ts
icetrust0212: Done

icetrust0212: Done

packages/verifiable-credentials/src/sharing-credential.ts
icetrust0212: Done

packages/verifiable-credentials/src/credentials.ts
icetrust0212: Done

packages/verifiable-credentials/src/sharing-credential.ts
aurelticot: You removed the default value on the wrong method

icetrust0212: Oh, yeah. Done

aurelticot (approved):

PR #440, opened 2024-05-31
Feature/430 vda xp reward client create

ITStar10:
Initial version of `vda-xp-reward-client`


packages/vda-xp-reward-client/src/blockchain/ownerApi.ts
```diff
@@ -0,0 +1,95 @@
...
+            throw new Error(`Failed to set conversion rate : ${response.reason}`);
+        }
+    }
+
+    /**
+     * @dev Withdraw tokens
+     * @param to - Recipient wallet address
+     * @param amount - Token amount to be withdrawn
+     */
+    public async withdraw(
```
tahpot, line 81: Do we need a `deposit()` method?

packages/vda-xp-reward-client/.env.example
```diff
@@ -0,0 +1,7 @@
+## These are used for test only.
+PRIVATE_KEY = "<Input private key of transaction sender>"
+RPC_URL="<Input RPC url for targeting chain>"
```
tahpot, line 3: We can remove this now as it should use default RPC URLs

tahpot (commented):
Looks good

packages/vda-xp-reward-client/src/blockchain/ownerApi.ts
ITStar10: In face, no need `deposit()` method.
The owner can deposit tokens by send the VDA token to the address of `XPReward` contract.

If we added `deposit()` function in the `XPReward` contract, the owner should approve the VDA token to the contract before calling the `deposit()` function. This involves 2 step for depositing.
I think, transferring VDA token to the `XPReward` contract is more conveninent.

packages/vda-xp-reward-client/.env.example
ITStar10: Just not clear.
`PRIVATE_KEY` is the private key of wallet that send the transaction.

And the `RPC_URL` is used in the tes script to manage the mock token.
image

packages/vda-xp-reward-client/src/blockchain/ownerApi.ts
tahpot: I'm proposing adding a `deposit()` method to this typescript library, not the smart contract.

A user of this typescript library needs a way to easily deposit tokens.

packages/vda-xp-reward-client/.env.example
tahpot: If the RPC_URL is only for the. test script, then make that clear. Add a comment in `.env.example` and maybe rename to `RPC_URL_TEST_DEPLOY`

ITStar10: There is comment in the `.env.example` file.
image

packages/vda-xp-reward-client/src/blockchain/ownerApi.ts
ITStar10: Added `depositToken()` function

PR #456, opened 2024-10-10
Extend Account implementation to pass signed sessions to node Account

aurelticot:


aurelticot:
I created a `SessionAccount` from scratch because the `LimitedAccount` and `ContextAccount` are extending `AutoAccount` which is very reliant on the `privateKey` property.

`SessionAccount` is more similar to `VaultAccount` because the session is from a context that already exists, so no need to create the context, link storage, unlink, etc.

tahpot (approved):

Issue #460, opened 2025-03-19
Support external wallets and social logins


aurelticot:
# Problem Statement

The current connection mechanism (Verida Connect) relies on the Verida Wallet mobile app as well as a (currently) centralised Auth server. The Verida Wallet is also the main way for users to create a Verida identity/account in the first place.

While having some purpose, the Verida Wallet adds friction to onboard new users.

A way we are exploring is to allow users create a Verida identity/account and connect to Verida via more usual methods such as email, social login (Google, Facebook, etc.), passkeys, as well as crypto wallets (Metamask, etc.).

Web3 now has a variety of "Wallet infrastructure" providers such as [Web3Auth](https://web3auth.io/), [ThirdWeb](https://thirdweb.com/), [Privy](https://www.privy.io/), [Reown](https://reown.com/) (WalletConnect) and others. They provide roughly the same capabilities in regards to onboarding users via the methods cited above. Technically they all eventually provide a crypto wallet and signer for the application integrating them.

This crypto wallet can do what the Verida Wallet is currently doing.

The Verida Wallet uses `AutoAccount` but as it has the private key available, it uses it directly, given the limited scoped of its actions.

By using an alternative wallet, we obviously won't have the private key but a signer instead.

# Original Assumption

The original assumption is that whatever wallet is created by the third-party wallet infrastructure, we would just use it to sign a hardcoded message and the signature would become the top level private key of the Verida Account, passed into `AutoAccount`. That top level private key is the same thing as the seed/private key generated in the Verida Wallet. Everytime the user would connect, they would sign the same hardcoded message to generate the top-level private key.

There is a problem with the existing Verida accounts from the Wallet though. We can't export their seed/PK and import them in Metamask & similar. Because the private key (that used to be available in the Verida Wallet and passed down the line in clear) will no longer be available. And if that Metamask is connected via the third-party and we follow the logic above (signing a hardcoded message becoming the top level private key), it will create a completely different Verida identity/account.

Rather than the original assumption described above, I wondered if we could use the wallet generated by the third-party as the actual controller of the Verida identity/account. We don't have access to the private key but we could use a `Signer` to sign and execute all necessary transactions necessary for the Verida protocol to operate.

But the SDK is critically relying on the private key in a few places (mostly `VdaDid` and `DIDdocument` apparently). 

Ideally, using a `Signer` rather than the PK directly would be the best solution (and can be backward compatible for `AutoAccount`), so making changes to the SDK is relevant despite the effort and regression risks.

[...]

PR #461, opened 2025-03-27
Support external wallets and social logins

aurelticot:


aurelticot:
On hold for now, see https://github.com/verida/verida-js/issues/460#issuecomment-2771574108

PR #465, opened 2025-04-14
Fix mismatch of legacy DID in `StorageLink.getLinks`

aurelticot:


tahpot (approved):
Good pickup. Approved.

